PR #17, opened 2020-07-16
First draft JSON Schema for Weed COCO

### LINKED ISSUE #20: expand schema yamls
Add schema yamls for all parts of the weedcoco blob.

hlydecker:
Does this yaml make sense? I have likely made numerous mistakes just extending on the original yaml draft developed for one AgContext item by @jnothman 

Closes #20


AgContext.yaml
```diff
@@ -0,0 +1,149 @@
+
+agcontext_name:
+  type:
+    anyOf:
+    - type: string
+    description: |-
+      Agricultural context name
+      string. Required.
```
jnothman, line 8: I'm not sure that we need a name. If it's required, I think its description should give some sense of what it's required for: will it be used as some rigid identifier.

AgContext.yaml
```diff
@@ -0,0 +1,149 @@
+
+agcontext_name:
```
jnothman, line 2: Since this is in AgContext.yaml, I don't think we need to prefix this by agcontext_

AgContext.yaml
```diff
@@ -0,0 +1,149 @@
...
+  type:
+    anyOf:
```
jnothman, line 4: you don't need anyOf when there's a single option. `type: string` would have been sufficient.

AgContext.yaml
```diff
@@ -0,0 +1,149 @@
...
+      A string representing the name of the agricultural context.
+
+crop_type:
+  type:
+    anyOf:
+    - type: string
+    description: |-
+      General crop type
+      string. Required.
+      A string describing the general cropping scenario (e.g. "root vegetable").
```
jnothman, line 18: If this is an arbitrary string (not from a controlled vocabulary), is it useful?

AgContext.yaml
```diff
@@ -0,0 +1,149 @@
...
+      - 5
+      - 6
+      - 7
+      - 8
+      - 9
+      - 99
+    description: |-
+      BBCH crop development stage
+      "na", "00", "0", "1", "2", "3", "4", "5", "6", "7", "8", "9", "99". Required.
+      One of eleven possible strings representing the BBCH phenological development stage of the crop. If this AgContext is not in a cropping environment, use value "na".
```
jnothman, line 41:
```suggestion
      One of eleven possible strings representing the approximate BBCH phenological development stage of the crop. If this AgContext is not in a cropping environment, use value "na".
```

AgContext.yaml
```diff
@@ -0,0 +1,149 @@
...
+
+location_continent:
```
jnothman, line 43: is this a useful granularity?

AgContext.yaml
```diff
@@ -0,0 +1,149 @@
...
+  type:
+    anyOf:
+    - type: string
```
jnothman, line 46: if useful it needs an enum

AgContext.yaml
```diff
@@ -0,0 +1,149 @@
...
+location_country:
+  type:
+    anyOf:
+    - type: string
+    description: |-
+      Country where AgContext is located
+      string. Required.
+      A string representing the country where the AgContext is located.
+
+location_state:
```
jnothman, line 61: Is this a useful granularity? Is there some other way of defining coarse-grained region? Even country + postcode? It's worth asking our horticulturalists what level of granularity is useful...

Or just request a lat-long with jitter of up to defined limits permitted for privacy reasons.

We are able to allow users to filter by region while using a more specific lat-long location in the data input and internally.

AgContext.yaml
```diff
@@ -0,0 +1,149 @@
...
+      State where AgContext is located
+      string. Required.
+      A string representing the state/province where the AgContext is located.
+
+upload_time:
+  type:
+    anyOf:
+    - type: DateTime<'Y-m-d H:i:s'>
+    description: |-
+      Data and time when this AgContext was uploaded.
```
jnothman, line 75:
```suggestion
      Date and time when this AgContext was uploaded.
```
uploaded to where?

I don't think this is user-input so it probably shouldn't be here. It might also make more sense to mark on a Collection than an AgContext?

AgContext.yaml
```diff
@@ -0,0 +1,149 @@
...
+      DateTime. Required.
+      System generated date time object recorded when a dataset with AgContext is uploaded.
```
jnothman, line 77: In UTC? Should we mark this with a Z in the timestamp?

AgContext.yaml
```diff
@@ -0,0 +1,149 @@
...
+  type:
+    anyOf:
+    - type: string
+    - type: string
+      enum:
+      - variable
+    description:
+      Type of camera
+      string.
+      A string describing the model/make of camera used for this AgContext. If this is not fixed within the AgContext, use value "variable".
```
jnothman, line 89: maybe we want this to be fixed within the agcontext, and force the user to create multiple agcontexts to describe multiple camera types??

AgContext.yaml
```diff
@@ -0,0 +1,149 @@
...
+      int or "variable". Required.
+      An integer representing the angle captured by the camera across the diagonal of an image, measured in degrees. If this is not fixed within the AgContext, use value "variable".
+
+emr_channels:
+  type:
+    anyOf:
+    - type: string
+      enum:
+      - visual
+      - NIR
```
jnothman, line 135: should we stick to lowercase? I'm not sure

AgContext.yaml
```diff
@@ -0,0 +1,149 @@
...
+      A string describing the electromagnetic radiation channels recorded by the camera.
+
+url:
+  type:
+    anyOf:
+    - type: string
+    description: |-
+      URL of AgContext
+      string.
+      A string recording the URL of the AgContext.
```
jnothman, line 149: I think this kind of stuff belongs in Collection, I think

AgContext.yaml
hlydecker: I realise now that this is just replicating the dataset/collection name. I don't think it is needed; as multiple agcontexts can exist within a dataset/collection.

AgContext.yaml
hlydecker: Currently it is arbitrary; I included this as a sort of place holder for a more controlled term that we could use - perhaps one from Guy/Michael?

AgContext.yaml
hlydecker: I really like that idea; it probably applies to all camera associated info.  camera_type for DeepWeeds is currently "weedlogger" which is Alex's name for his combination of lens, sensor, and computer. However for CWFID, I used the model name of the camera that they used. Should we also be including objects for camera_lens, etc.?

AgContext.yaml
hlydecker: This section would likely benefit from controlled vocabulary. for a moment I considered having wavelengths

AgContext.yaml
```diff
@@ -0,0 +1,157 @@
...
+      string
+      description: |-
+        General crop type
+        str. Required.
+        A string describing the general cropping scenario (e.g. "root vegetable").
+
+  crop_stage:
+    type:
+      string
+        enum:
```
jnothman, line 38: the indentation here is also broken.

Maybe you'd be more comfortable working in JSON than YAML, since it's clearer what lines up to what in JSON. Everyone else uses JSOn for JSON Schemas. I just find YAML a bit more readable (and less finicky about commas)

jnothman:
I decided I'd better try out some tools to validate this as correct JSON Schema, and to see if existing tools can produce friendly HTML documentation. (There're also tools to turn JSON schema into a data entry form, which, given our conversation yesterday, might be more familiar to the researchers.)

AgContext.yaml
hlydecker: JSON's bracket and whitespace heavy nature make it pretty easy for a novice like me to pick up. FYI this is my first time really doing much with YAML, and I've mostly been proceeding following Atom's queues re formatting.

jnothman: So convert it to json and continue there!

jnothman:
I've learnt several things:
* I can't get [doca](https://github.com/cloudflare/json-schema-tools/tree/master/workspaces/doca) to work.
* After converting the YAML to JSON, I get no errors when trying to use the Python jsonschema tool: 
  ```py
  import json, jsonschema, requests
  meta_schema = requests.get('https://json-schema.org/draft/2019-09/schema').json()
  our_schema = json.load(open("/Users/joel/repos/sih/weed-id-grdc/Weed-ID-Interchange/tmp-schema/AgContext.json"))
  jsonschema.validate(instance=obj, schema=meta_schema)
  ```
  passes without error.
* Using https://www.jsonschemavalidator.net/ shows appropriate errors. We should find a command-line tool with similar functionality. Once we've identified such a tool, it should be straightforward to create a github workflow to check our work for us.
* `int` must be changed to `number`
* I got [jsonschema2md](https://github.com/adobe/jsonschema2md) to work, but I needed the filename to have suffix `.schema.json`.

jnothman:
Here's how things render with jsonschema2md:
image

jnothman:
And I've bashed my head against a whole series of command-line validators. ajv and pajv both tell me that my schema is invalid. yajsv doesn't appear on my path after installation.

Oh no. It seems ajv works, but not for the latest version of JSON Schema. With JSON Schema 07, it works fine... 

But we only really need the github workflow, plus to download the appropriate json schema meta-schema:
```
curl -L https://json-schema.org/draft-07/schema > /tmp/jsonschema07.schema
```
then run the workflow at https://github.com/marketplace/actions/validate-json with 
```
          INPUT_SCHEMA: /tmp/jsonschema07.json
          INPUT_JSONS: AgContext.json,Collection.json,...
```
Not sure if we can use a glob to match all INPUT_JSONS files in a directory.

AgContext.yaml
```diff
@@ -0,0 +1,157 @@
...
+        If this AgContext is not in a cropping environment, use value "na".
+
+  weather_description:
+    type:
+      string
+      description:
+        Description of weather when images were captured.
+        str.
+
+  location_continent:
```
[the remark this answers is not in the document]
hlydecker: For now I am going to specify these in decimal degrees using EOSG 4326 projection

Images.yaml
```diff
@@ -0,0 +1,56 @@
...
+    description: |-
+      A number identifying the image.
+
+  file_name:
+    type: string
+    description: |-
+      A string identifying the filename of the image.
+      This string does not include the path of the image.
+
+    license:
```
jnothman, line 24: why is this and below nested within file_name? Please outdent.

Images.yaml
```diff
@@ -0,0 +1,56 @@
...
+      type: number
+      minimum: 1
+      description: |-
+        Vertical pixel count.
+
+    resolution:
+      type: number
+      minimum: 1
+      description: |-
+        Image resolution, in pixels.
```
jnothman, line 50: Is this not already determined from width and height? is this required in COCO?

Images.yaml
```diff
@@ -0,0 +1,56 @@
...
+  id:
+    type: number
+    description: |-
+      A number identifying the image.
+
+  file_name:
+    type: string
+    description: |-
+      A string identifying the filename of the image.
+      This string does not include the path of the image.
```
jnothman, line 22: I assume it can include the path of the image relative to the image base?

Or could it be a URL in the general case?

annotations_categories.yaml
```diff
@@ -0,0 +1,26 @@
+allOf:
+  - type: object
+    properties:
+
+  - anyOf:
```
jnothman, line 5: I think this implies that the file should be annotations.yaml, not annotations_categories.yaml. It should be supporting multiple variants.

.github/workflows/validate_json.yml
```diff
@@ -16,8 +16,13 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
+      - name: Compile yaml schema to JSON
+        run: |
+          pip install pyyaml
+          python -c 'import json, yaml, sys; json.dump(yaml.safe_load(sys.stdin), sys.stdout, indent=2)' < AgContext.yaml > AgContext.json
```
jnothman, line 22: This should be extended to cover all the YAML files.

There should be a main schema file for the overall COCO document, and we can use that in the validator below. Let me know if I can support you in doing this.

AgContext.yaml
```diff
@@ -0,0 +1,321 @@
...
+      - cotton
+      - black_plastic
+      - white_plastic
+      - woodchips
+      - other
+    description: |-
+      Surface cover type.
+      One of several strings describing the background cover that is behind any plants in the images.
+
+  surface_coverage:
```
hlydecker, line 224: Would this be better represented as a set of possible number ranges?

AgContext.yaml
```diff
@@ -0,0 +1,321 @@
...
+      - ripening or maturity of fruit and seed
+      - senescence
+      - beginning of dormancy
+      - na
+    description: |-
+        BBCG descriptive text.
+        One of several possible strings describing the stage of the crop, chosen from a list of possible terms used by the BBCH.
+        If this AgContext is not in a cropping environment, use value "na".
+
+  bbch_code:
```
hlydecker, line 84: This is a bit silly looking; would something like a regex be possible here?

AgContext.yaml
```diff
@@ -0,0 +1,321 @@
...
+        A number identifying the AgContext.
+
+  crop_type:
+    type:
+      string
+    description: |-
+          General crop type.
+          A string describing the general cropping scenario (e.g. "root vegetable").
+
+  grains_descriptive_text:
```
hlydecker, line 36: I currently have the three different crop stage descriptions in here. Is it possible to link them to each other, or is that over engineering for the schema?

AgContext.yaml
jnothman: A regex is certainly possible. https://json-schema.org/understanding-json-schema/reference/regular_expressions.html `pattern: "^gs[0-9][0-9]$|^na$`. Although this includes gs00 which is currently not permitted here. (We can exclude that from the pattern if necessary.)

AgContext.yaml
```diff
@@ -0,0 +1,321 @@
...
+      - milky_dough
+      - dough
+      - ripening
+      - na
+    description: |-
+        Grains descriptive text.
+        One of ten possible strings describing the crop developmental stage.
+        If this AgContext is not in a cropping environment, use value "na".
+
+  bbch_descriptive_text:
```
jnothman, line 56: I think we will have this available in a data entry form, but shouldn't need it as a separate property in the JSON schema. I don't see anything wrong with having a `value_descriptions` or `bbch_descriptions` in the definition of `bbch_code` that could be used to construct data entry forms or search interfaces... But maybe we're getting too generic then and extending JSON schema beyond its purpose.

AgContext.yaml
jnothman: As per bbch_descriptive_text, I don't think these should belong here.

Ideally we'd get to the stage of mocking up a search form soon. Maybe you and I should discuss how we'd get there.

AgContext.yaml
jnothman: I'm not sure what you mean by "a set of possible number ranges". Do you mean that these would be "type: number" in JSON schema?

collections.yaml
```diff
@@ -0,0 +1,54 @@
...
+    accrual_policy:
+      type: string
+      enum:
+        - closed
+        - open
+      description: |-
+        A description of the accrual policy for this collection.
+        Open collections can be contributed to, and closed collections cannot be.
+
+  - type: object
```
jnothman, line 40: Is this to allow for a Collection to either be a DCMI-complete Collection definition, or to be a "subset of a collection" by reference?

If so:
* allOf above should be anyOf, since it's an either/or situation.
* I'm not sure why we want `image_id` here.
* Why do you prefer this to having `collection_memberships`?

jnothman (commented):
more things I saw

schema tools.ipynb
```diff
@@ -0,0 +1,367 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": 2,
+   "metadata": {},
+   "outputs": [
```
jnothman, line 7: Let's clear the output.

Let's just put this in a .py script

hlydecker: Done.

AgContext.yaml
```diff
@@ -0,0 +1,269 @@
...
+      - legume
+      - cotton
+      - black_plastic
+      - white_plastic
+      - woodchips
+      - other
+      - none
+    description: |-
+      Surface cover type.
+      One of several strings describing the background cover that is behind any plants in the images.
```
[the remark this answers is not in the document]
hlydecker: Good catch, that is a less ambiguous way to put it.

AgContext.yaml
```diff
@@ -0,0 +1,269 @@
...
+  surface_coverage:
+    type: string
+    enum:
+      - "0-25"
+      - "25-50"
+      - "50-75"
+      - "75-100"
+    description: |-
+      Percent of coverage in image.
+      Approximate measurement of the percent of the soil in the image that is covered by the surface cover.
```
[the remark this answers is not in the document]
hlydecker: I agree that this sounds like it may be of questionable utility; I had clarified this with Guy and it is intended to be a rough estimate for the entire dataset.

AgContext.yaml
```diff
@@ -0,0 +1,269 @@
...
+        Approximate latitude location of the AgContext, in decimal degrees.
+
+  location_long:
+    type:
+      number
+    description:
+        Longitude in decimal degrees.
+        Approximate longitude location of the AgContext, in decimal degrees.
+
+  location_datum:
```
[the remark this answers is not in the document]
hlydecker: I think it is sensible to require a specific projection for the spatial coordinates. We can check what would be the best for ag people, but I imagine 4326 should work.

PR #117, opened 2020-10-09
Segmentation mask convertor

hlydecker:
WIP. Some questions need answering.

- Currently only works with hard coded categories, for single category images. Will need to adapt it to link categories to masks by colour codes.

- Currently is missing license and info objects.


hlydecker:
@jnothman this has been built based on working with the carrots images/masks. These are binary (black = nothing, white = plant), and there are usually one or two main objects with a bunch of super tiny ones, created as a result of the masking/annotation process. How should we deal with these? Subtract the tiny polygons, or potentially subsume them with the main ones?

jnothman:
> this has been built based on working with the carrots images/masks

Do you mean ginger?
Why not add a test image and test mask to the repo, and design a test case?

>  These are binary (black = nothing, white = plant),

I had imagined a config file `color-category-map.yml`:
```py
FFFFFF: "weed: UNSPECIFIED"
```
This is extensible to multiple colours.

> Subtract the tiny polygons, or potentially subsume them with the main ones?

I don't think so. We should be authentic to the input. On this, it's not our job to be opinionated.

COCO assumes that there are multiple (or, more precisely, one or more) polygons. #90 codifies this in the schema: https://github.com/Sydney-Informatics-Hub/Weed-ID-Interchange/blob/6ef3a168215627c039b74224a73f2782a98a4b63/weedcoco/schema/Annotation.yaml#L33-L42.

Note that an alternative representation is as a mask (a 2d binary array) encoded with RLE and special encoding that only seems to be handled by COCO API (https://github.com/cocodataset/cocoapi/blob/8c9bcc3cf640524c4c20a9c40e89cb6a2f2fa0e9/common/maskApi.c#L204-L231).

Turning the image into a mask, based on known annotation colours, and then using pycocotools, may provide more straightforward solutions than thresholding and opencv, which is designed more for photography than discrete masking.

hlydecker:
I have no idea why I wrote carrots; yes I meant ginger!!!

weedcoco/importers/mask.py
```diff
@@ -0,0 +1,117 @@
...
+from weedcoco.utils import add_collection_from_file
+
+# TODO: extend functionality to read categories from a file
+categories = [
+    {"id": 0, "name": "crop: zingiber officnale"},
+]
+
+
+def generate_masks_contours(mask_path):
+
```
jnothman, line 25: This blank line violates PEP257. I'm surprised black lets it through

weedcoco/importers/mask.py
```diff
@@ -0,0 +1,117 @@
...
+        'categories'.
+        Other COCO components should be added as a post-process.
+    """
+
+    images = []
+    annotations = []
+
+    image_id = 0
+    for filename in os.listdir(image_dir):
+        if filename.endswith(".png") or filename.endswith(".jpg"):
```
jnothman, line 62: there should be an else clause that warns or raises an error if the file type is unexpected

jnothman: `jpeg` and `tif` might also be possible extensions

jnothman: A neat shorthand
```suggestion
        if filename.endswith((".png", ".jpg", ".jpeg", ".tif", ".tiff")):
```

weedcoco/importers/mask.py
```diff
@@ -0,0 +1,117 @@
...
+                "id": image_id,
+                "file_name": filename,
+            }  # TODO: use get_image_dimensions
+            images.append(coco_image)
+
+            annotation = {
+                "id": image_id,
+                "image_id": image_id,
+                "category_id": 0,
+                "segmentation": generate_masks_contours(str(mask_dir / filename)),
```
jnothman, line 73: the segmentation mask should never be jpg. I expect it should always be png. jpg is a lossy format designed for photos; png is designed for lossless compression, i.e. for exact replication of the input mask. jpg will introduce an arbitrary set of colours between the mask colours.

weedcoco/tests/importers/mask_data/agcontext-ginger.yaml
```diff
@@ -0,0 +1,268 @@
+### crop_type ###
+# description: 'Crop type.
+
+#   One of several strings describing the crop grown in the image.'
```
jnothman, line 4: these comments aren't needed in a test file. You can do `sed -E 's/^$|^#/d'` on the file to get only the content lines

jnothman (commented):
I don't think coco_from_mask.json should be included in the repo. Rather:
* there should be at least one pytest test case checking that the converter works;
* we might add a script `search/scripts/index_rds_images.sh` which is given the path to the RDS root, and converts and loads data from there.

hlydecker:
Major changes are afoot. Using this [blog post](https://www.immersivelimit.com/create-coco-annotations-from-scratch) as a template, to redesign this convertor to work with colour category mapping. In some ways this is reinventing aspects of the opencv contour generator, but it may be a better direction for our use case.

weedcoco/tests/importers/mask_data/category_name_chobbitty
```diff
@@ -0,0 +1,3 @@
+FF0000: "weed: broadleaf"
```
jnothman, line 1: call this category_name_cobbity.yaml and you can load it with weedcoco.utils.load_json_or_yaml

weedcoco/importers/masks2.py
```diff
@@ -0,0 +1,125 @@
+from PIL import Image, ImageDraw
+import numpy as np
+from skimage import measure
+from shapely.geometry import Polygon, MultiPolygon
+
+
+def create_sub_masks(mask_image):
```
jnothman, line 7: I suspect you're overengineering this.
You should just convert the image bitmap into an array:
```py
im = PIL.Image.open(...)
arr = np.asarray(im)
```
arr is of shape (height, width, depth), where depth=3 for the cobbity masks.
You can see that there are three distinct colours with:
```py
distinct_colours = np.unique(np.reshape(arr, (-1, 3)), axis=0)
```
You can extract out the first of those colours (black) with:
```py
mask = (arr == distinct_colours[0]).all(axis=-1)
```

Thus:
image

weedcoco/importers/masks2.py
```diff
@@ -0,0 +1,125 @@
...
+    return sub_masks
+
+
+def create_sub_mask_annotation(
+    sub_mask, image_id, category_id, annotation_id, is_crowd
+):
+    # Find contours (boundary lines) around each sub-mask
+    # Note: there could be multiple contours if the object
+    # is partially occluded. (E.g. an elephant behind a tree)
+    contours = measure.find_contours(sub_mask, 0.5, positive_orientation="low")
```
jnothman, line 41: Once you've got a boolean mask array, why not just use pycocotools.mask.encode to get a valid COCO segmentation out?
image

hlydecker: Ok you weren't kidding that there was a much simpler issue. I think my attempt to make it simpler ended up even further over engineering it...

jnothman:
I'm not yet happy with the sufficiency of the tests, but this is otherwise ready for review.

jnothman:
Maybe I should open a new PR so that Henry can review. @hlydecker would you like to and are you available to do so?

hlydecker:
Happy to review this sometime this afternoon!

hlydecker (commented):
This looks good and the existing tests are sensible enough. Most of my comments are just some suggestions for making warnings and errors more clear to potential users.

I do wonder what sort of other tests could be included. test_basic is indeed basic but it does test the basic functionality!

hlydecker:
Ah of course now I realise the awkwardness here; it makes sense that I cannot be a reviewer for my own pull request even if the actual content is not my progeny.

jnothman:
Want to check out the changes since last review, @hlydecker and gimme a tick if possible?

hlydecker:
Will take a look this afternoon!

weedcoco/importers/mask.py
```diff
@@ -30,7 +30,7 @@ def generate_segmentations(mask_path, color_map, colors_not_found):
     Yields
     ------
     segmentation : str
-        COCO segmentation string
+        COCO segmentation string in compressed RLE format
```
hlydecker, line 33: Good addition to the documentation

weedcoco/tests/importers/test_mask.py
```diff
@@ -5,6 +5,13 @@
 
 from weedcoco.importers.mask import main
 
+# TODO: Further testing plan:
+# * check segmentation RLE string can be read back in and reproduces the mask
+# * check handling of missing correspondence between mask and image files
+# * check handling of different image file formats
+# * check handling of agcontext
```
hlydecker, line 12: This would be good. The agcontext ingestion + testing would probably make sense as something that is shared as a utility called by each individual converter.

weedcoco/importers/mask.py
```diff
@@ -163,7 +163,7 @@ def _image_name_to_mask(name):
         warnings.warn(
             f"{len(categories)} categories defined, but only "
             f"{len(categories_found)} of these are present in masks. "
-            f"Missing are {missing_category_colors}"
+            f"These categories were not found: {missing_category_colors}"
```
hlydecker, line 166: Both this change and the one at 118 are great improvements in the clarity of the messages to users.

hlydecker (commented):
LGTM; not much was changed. Documentation changes have improved clarity.

The testing plan sounds good to me as well.

weedcoco/importers/mask.py
jnothman: > Both this change and the one at 118 are great improvements in the clarity of the messages to users.

They were both your explicit contributions! :D

jnothman:
Okay to merge as is, despite TODOs?

hlydecker:
I'd say so :)

jnothman:
Thanks for the review @hlydecker

PR #137, opened 2020-10-27
Add deposit/repository functionality

ElevnLi:
I've set some validation and constraints, and will work on tests.


weedcoco/utils.py
```diff
@@ -50,3 +51,8 @@ def add_collection_from_file(coco, collection_path):
         for annotation in coco["annotations"]
     ]
     return coco
+
+
+def get_image_average_hash(path):
+    """Return an average hash of an image"""
+    return str(imagehash.average_hash(PIL.Image.open(path)))
```
jnothman, line 58: Interesting. I had thought we'd use a cryptographic hash – to avoid byte-identical images – rather than a locality-sensitive hash. I don't know if it will be overly sensitive, when this algorithm was, I think, designed for finding similar images, and for showing them to a person to confirm.

weedcoco/importers/deposit.py
```diff
@@ -0,0 +1,107 @@
+import argparse
```
jnothman, line 1: Deposit is not an `importer`. Maybe it should be `weedcoco/repo/deposit.py` since it is now about a repository rather than the schema.

weedcoco/importers/deposit.py
```diff
@@ -0,0 +1,107 @@
...
+    ap.add_argument("--image-dir", default="cwfid_images", type=pathlib.Path)
+    ap.add_argument("--repository-dir", default="repository", type=pathlib.Path)
+    args = ap.parse_args()
+
+    image_hash = create_image_hash(args.image_dir)
+    validate_duplicate_images(image_hash)
+    validate_existing_images(args.repository_dir, image_hash)
+
+    dataset_dir = setup_dataset_dir(args.repository_dir)
+    generate_weedcoco(args.weedcoco_path, dataset_dir, image_hash)
```
jnothman, line 102: "generate" seems strange. Maybe "deposit"? Another awkward thing is that ideally `validate` accepts the `image_dir` as well, to check that annotations match the images. So we will need to pass that in.

weedcoco/importers/deposit.py
```diff
@@ -0,0 +1,107 @@
...
+    for image_origin in os.listdir(raw_dir):
+        image_path = dataset_dir / ("images/" + image_hash[image_origin])
+        if not os.path.isfile(image_path):
+            copyfile(raw_dir / image_origin, image_path)
+
+
+def generate_weedcoco(weedcoco_path, dataset_dir, image_hash):
+    with open(weedcoco_path) as f:
+        weedcoco = json.load(f)
+    validate(weedcoco)
```
jnothman, line 78: it might be appropriate to do the validation after the modification. We wouldn't want to inadvertently modify it in a way that broke the format.

weedcoco/importers/deposit.py
```diff
@@ -0,0 +1,107 @@
...
+    ):
+        raise ValidationError("There are identical images in the repository.")
+
+
+def get_all_existing_hash(repository_dir):
+    return reduce(
+        lambda acc, it: acc.union(set(it[2])),
+        filter(lambda dir: dir[0].endswith("/images"), os.walk(repository_dir)),
+        set(),
+    )
```
jnothman, line 31: I personally find following much more readable:
```suggestion
    return {
        filename
        for dirpath, dirnames, filenames
        if dirpath.endswith("/images")
        for filename in filenames
    }
```
I note two things about `.endswith("/images")`:
1. This won't work on Windows. Using `basename()` would be more cross-platform.
2. If we use a directory structure (e.g. using the first two characters of the image hash as a directory name) to split up large directories of images, then you'll need a different filter.

weedcoco/importers/deposit.py
```diff
@@ -0,0 +1,107 @@
...
+    os.mkdir(dataset_dir)
+    return dataset_dir
+
+
+def create_image_hash(image_dir):
+    return {
+        image_name: f'{get_image_average_hash(image_dir / image_name)}.{image_name.split(".")[-1]}'
+        for image_name in os.listdir(image_dir)
+        if image_name.lower().endswith(
+            (".png", ".jpg", ".jpeg", ".tiff", ".bmp", ".gif")
```
jnothman, line 59: I hope we're not dealing with bmp and gif! :D If we are, we should be converting them to png!

weedcoco/importers/deposit.py
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+
+def migrate_images(dataset_dir, raw_dir, image_hash):
+    image_dir = dataset_dir / "images"
+    if not os.path.isdir(image_dir):
+        os.mkdir(image_dir)
+
+    for image_origin in os.listdir(raw_dir):
+        image_path = dataset_dir / ("images/" + image_hash[image_origin])
```
jnothman, line 70: using pathlib.Path:
```suggestion
        image_path = dataset_dir / "images" / image_hash[image_origin]
```

weedcoco/importers/deposit.py
```diff
@@ -0,0 +1,107 @@
...
+            dataset_dir = repository_dir / f"dataset_{latest_dataset_dir_index + 1}"
+        else:
+            dataset_dir = repository_dir / "dataset_0"
+    os.mkdir(dataset_dir)
+    return dataset_dir
+
+
+def create_image_hash(image_dir):
+    return {
+        image_name: f'{get_image_average_hash(image_dir / image_name)}.{image_name.split(".")[-1]}'
```
jnothman, line 56: `image_name.split(".")` should really be `image_name.rsplit(".", 1)` in case a path has multiple `.`s in it. Rather, can use
```suggestion
        image_name: f'{get_image_average_hash(image_dir / image_name)}{os.path.splitext(image_name)[-1]}'
```

weedcoco/importers/deposit.py
jnothman: I think this function has forgotten to remove the filename extension.

weedcoco/importers/deposit.py
```diff
@@ -0,0 +1,107 @@
...
+    )
+
+
+def setup_dataset_dir(repository_dir):
+    if not os.path.isdir(repository_dir):
+        os.mkdir(repository_dir)
+        dataset_dir = repository_dir / "dataset_0"
+    else:
+        if "dataset_0" in os.listdir(repository_dir):
+            latest_dataset_dir_index = max(
```
jnothman, line 40: Instead of the `if`-`else`, just use `max([...], default=0)`?

jnothman (commented):
Some comments! Thanks :)

weedcoco/utils.py
jnothman: It would be good to check that there are no false positive hash identities on RDS at the moment, or images that are very close in hash. That would mean that we should increase the hash size.

ElevnLi: There are [several ways](https://fullstackml.com/wavelet-image-hash-in-python-3504fdd282b5) to hash that library offers, though I think any of them should be sufficient as the identifier. If we want to make it more precise, we can combine some of these hashes, which will cost more time in processing as well.

weedcoco/importers/deposit.py
ElevnLi:
```python
return {
        os.path.splitext(filename)[0]
        for dirpath, dirnames, filenames in os.walk(repository_dir)
        if os.path.basename(dirpath) == "images"
        for filename in filenames
    }
```

weedcoco/utils.py
jnothman: The different hashes should have different *types* of false positive, but the main way to reduce false positives for LSH is to increase the hash size.

weedcoco/importers/deposit.py
jnothman: Looks more precise, and readable

weedcoco/utils.py
ElevnLi: The default hashsize is 8. I increased it to 16 in the latest commit. A sample file name can be `0b5c1a7c18fc49f80dfc04fc0cfc41f901f800d82018009c019c09bc033c017c.png`

weedcoco/utils.py
```diff
@@ -55,4 +55,4 @@ def add_collection_from_file(coco, collection_path):
 
 def get_image_average_hash(path):
     """Return an average hash of an image"""
-    return str(imagehash.average_hash(PIL.Image.open(path)))
+    return str(imagehash.average_hash(PIL.Image.open(path), hash_size=16))
```
jnothman, line 58: This will certainly reduce the rate of false positives, but it would be good to substantiate this by doing an analysis of the similarity of existing images. That involves basically doing:

```
X = np.asarray([imagehash.average_hash(PIL.Image.open(path)).hash.flatten()
                for path in all_image_paths_on_RDS)
distances = scipy.spatial.pdist(X, metric='hamming')
```
and then looking at the histogram/distribution of `distances`. Alternatively, plotting the seaborn.clustermap of X with hamming distance might be demonstrative.

ElevnLi: The distributions are as below:
Hash size = 8
RDS_distribution_hashsize_8
Hash size = 16
RDS_distribution_hashsize_16
There are some directories where images are mainly black, which contribute the high similarity, but still there no identical images with both sizes of hashing.

jnothman: Excellent. Yes, I thought the segmentation masks would be the source of clashes. Happy to adopt a moderate hash, e.g. hash_size=10 (in practice that's 50% more specific a hash).

jnothman: Thanks for that, by the way! Makes me a bit more comfortable.

jnothman:
`weedcoco/tests/repo/deposit_data/basic_1/images` hasn't been committed, which is why the tests are failing.

jnothman:
Should this still be regarded as in draft? Are you happy with the tests?

weedcoco/repo/deposit.py
```diff
@@ -0,0 +1,110 @@
...
+    if len(all_existing_hash.union(addition_hash)) != len(all_existing_hash) + len(
+        addition_hash
+    ):
+        raise ValidationError("There are identical images in the repository.")
+
+
+def get_hashset_from_image_name(image_hash):
+    return set(
+        map(lambda image_name: os.path.splitext(image_name)[0], image_hash.values())
+    )
```
jnothman, line 28: I think it is more pythonic, and more readable, to express this as
```suggestion
    return {os.path.splitext(image_name)[0] for image_name in image_hash.values()}
```

weedcoco/repo/deposit.py
```diff
@@ -0,0 +1,110 @@
...
+    ap.add_argument("--repository-dir", default="repository", type=pathlib.Path)
+    args = ap.parse_args(args)
+
+    image_hash = create_image_hash(args.image_dir)
+    validate_duplicate_images(image_hash)
+    validate_existing_images(args.repository_dir, image_hash)
+
+    dataset_dir = setup_dataset_dir(args.repository_dir)
+    deposit_weedcoco(args.weedcoco_path, dataset_dir, args.image_dir, image_hash)
+    migrate_images(dataset_dir, args.image_dir, image_hash)
```
jnothman, line 106: Let's bundle this all into a single deposit function that can be called/tested from Python.

weedcoco/tests/repo/test_deposit.py
```diff
@@ -0,0 +1,124 @@
...
+        if isinstance(obj, dict):
+            return sorted((k, ordered_json(v)) for k, v in obj.items())
+        if isinstance(obj, list):
+            return sorted(ordered_json(x) for x in obj)
+        else:
+            return obj
+
+    return all(
+        [
+            ordered_json(json.loads(open(dir1 / dir / "weedcoco.json").read()))
```
jnothman, line 74: JSON can be parsed from a stream when not in a string:
```suggestion
            ordered_json(json.load(open(dir1 / dir / "weedcoco.json")))
```

weedcoco/tests/repo/test_deposit.py
```diff
@@ -0,0 +1,124 @@
...
+                repository_dir,
+            ]
+            args = [str(arg) for arg in args]
+            main(args)
+            return test_repo_dir
+
+    return Executor()
+
+
+def _cmp_files(dir1, dir2):
```
jnothman, line 39: A problem with this design is that when it fails, all you'll get it "True" or "False". That's going to be a big pain to debug. Renaming these to `assert_files_equal` and `assert_weedcoco_equal`, and correspondingly making them consist of a series of `assert` statements, will mean that pytest can do its best to show you a rich analysis of the expression causing the failure: it can show you the parts of dicts that are not equal in `assert actual_dict == expected_dict`, or the values of variables in a comparison expression (and other local variables).

weedcoco/tests/repo/deposit_data/basic_1/weedcoco.json
```diff
@@ -0,0 +1,732 @@
+{
+    "images": [
+        {
+            "id": 1,
+            "file_name": "/Users/elevn/Documents/workspace/Weed-ID-Interchange/weedcoco/tests/repo/deposit_data/basic_1/images/001_image.png",
```
jnothman, line 5: Hmm... I had expected input to have relative paths, i.e. just filenames.

I've not reviewed how you currently handle absolute paths. I think we can assume that all image files are in one directory, and just use the basename to reference them in a collection of uploaded images.

jnothman (commented):
On top of these comments, I don't think we benefit from having a "complete" dataset in here. It's too big to burden developers with having the weedcoco in the repository, let alone the images. We should probably also make sure that we're only including a small proportion of an existing dataset for copyright reasons. So if we can cut that back easily, that would be great.

I'll open a separate issue about prefixing directories in the repo by the first two chars of the image hash.

ElevnLi:
> It's too big to burden developers with having the weedcoco in the repository, let alone the images.

So I just remove everything for complete test right?

jnothman:
> So I just remove everything for complete test right?

If size is the only difference between this and the other tests, then yes, please.

weedcoco/index/indexing.py
```diff
@@ -0,0 +1,140 @@
+import argparse
+import os
+import pathlib
+import json
+import requests
+from math import ceil
+
+
+class ElasticSearchIndex:
```
jnothman, line 9: I had imagined what is now in weedcoco/index to live with the webapp stuff... It's about the webapp, not about the format or repository... but I suppose it can live together? It's a bit confusing, because the celery stuff in django will need to call this.

weedcoco/tests/repo/test_deposit.py
```diff
@@ -0,0 +1,106 @@
...
+            return obj
+
+    return all(
+        [
+            ordered_json(json.load(open(dir1 / dir / "weedcoco.json")))
+            == ordered_json(json.load(open(dir2 / dir / "weedcoco.json")))
+            for dir in os.listdir(dir2)
+            if re.fullmatch(r"^dataset_\d+$", dir)
+        ]
+    )
```
jnothman, line 72: Can we just do
```suggestion
    for dir in os.listdir(dir2):
        if re.fullmatch(r"^dataset_\d+$", dir):
            assert json.load(open(dir1 / dir / "weedcoco.json")) == json.load(open(dir2 / dir / "weedcoco.json"))
```
?

ElevnLi: In this way we need to assume everything inside the image list should be in the same order. For example in the following case, `a == b` will be False.
```python
import json

a = json.loads("""
{
    "errors": [
        {"error": "invalid", "field": "email"},
        {"error": "required", "field": "name"}
    ],
    "success": false
}
""")

b = json.loads("""
{
    "success": false,
    "errors": [
        {"error": "required", "field": "name"},
        {"error": "invalid", "field": "email"}
    ]
}
""")
```

weedcoco/index/indexing.py
ElevnLi: Yeah that's the goal. I'll make adjustments if the current structure can't be called by celery and django appropriately.

weedcoco/tests/repo/test_deposit.py
jnothman: The assertion will work fine for changes in dict order, but not list order. List order is usually meaningful, so I wouldn't want it to be permissive for all lists. I'd rather if we made sure that the error output was deterministic than make the assertion too lenient.

weedcoco/index/indexing.py
```diff
@@ -0,0 +1,140 @@
...
+                # determine available task types
+                image["task_type"].add("classification")
+                if "segmentation" in annotation:
+                    image["task_type"].add("segmentation")
+                    image["task_type"].add("bounding box")
+                if "bbox" in annotation:
+                    image["task_type"].add("bounding box")
+            image["task_type"] = sorted(image["task_type"])
+
+        self.coco = coco
```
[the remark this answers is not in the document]
jnothman: It might make sense to only store the images... Or even a generator of images.

weedcoco/index/indexing.py
```diff
@@ -0,0 +1,140 @@
...
+
+    def generate_batches(self):
+
+        coco = self.coco
+        if "images" in coco and len(coco["images"]) > 0:
+            for i in range(ceil(len(coco["images"]) / self.index_batch)):
+                with open(f"{self.temp_index_dir}/index{str(i+1)}.json", "w") as f:
```
jnothman, line 105: Are we writing to disk for debugging purposes, or some other reason? It seems strange to write to disk if we're only going to read it back in again right away, and into memory (rather than streaming from disk).

weedcoco/index/indexing.py
```diff
@@ -0,0 +1,140 @@
...
+                    for image in coco["images"][
+                        i
+                        * self.index_batch : min(
+                            len(coco["images"]), (i + 1) * self.index_batch
+                        )
+                    ]:
+                        json.dump({"index": {"_index": "weedid", "_type": "image"}}, f)
```
jnothman, line 112: we should make the index name a config parameter of the class, IMO

weedcoco/index/indexing.py
```diff
@@ -0,0 +1,140 @@
...
+        thumbnail_dir,
+        temp_index_dir="temp",
+        index_batch=30,
+        es_url="http://localhost:9200/",
+        coco={},
+    ):
+        self.weedcoco_path = weedcoco_path
+        self.thumbnail_dir = thumbnail_dir
+        self.temp_index_dir = temp_index_dir
+        self.index_batch = index_batch
```
jnothman, line 22: Might be better named `batch_size`

weedcoco/index/indexing.py
```diff
@@ -0,0 +1,140 @@
...
+
+        self.coco = coco
+
+    def generate_batches(self):
+
+        coco = self.coco
+        if "images" in coco and len(coco["images"]) > 0:
+            for i in range(ceil(len(coco["images"]) / self.index_batch)):
+                with open(f"{self.temp_index_dir}/index{str(i+1)}.json", "w") as f:
+                    for image in coco["images"][
```
jnothman, line 106: One idiom for slicing batches from a list looks like:
```py
for start in range(0, len(images), self.index_batch):
    batch = images[start:start+self.index_batch]
```

weedcoco/index/indexing.py
jnothman: We could just generate a string for each batch:
```py
for batch in image_batches:
    blobs = []
    for image in batch:
        blobs.append('{"index": {"_index": "weedid", "_type": "image"}}')
        blobs.append(json.dumps(out))
    yield "\n".join(blobs)
```
In this case the post method would take an iterable of strings to post.

I note that the [Python `elasticsearch` package](https://pypi.org/project/elasticsearch/) manages the batching of objects for bulk upload, but also uses multithreading to post them more efficiently.

weedcoco/index/indexing.py
```diff
@@ -0,0 +1,140 @@
...
+                        f.write("\n")
+
+    def post_index(self):
+
+        files = [
+            file for file in os.listdir(self.temp_index_dir) if file.endswith(".json")
+        ]
+        for item in files:
+            requests.post(
+                f"{self.es_url}_bulk",
```
jnothman, line 124: safer to do:
```suggestion
                urllib.parse.urljoin(self.es_url), "_bulk"),
```

weedcoco/index/indexing.py
```diff
@@ -0,0 +1,142 @@
...
+                        * self.index_batch : min(
+                            len(coco["images"]), (i + 1) * self.index_batch
+                        )
+                    ]:
+                        json.dump({"index": {"_index": "weedid", "_type": "image"}}, f)
+                        f.write("\n")
+                        json.dump(image, f)
+                        f.write("\n")
+
+    def post_index(self):
```
jnothman, line 117:
```suggestion
    def post_to_index(self):
```
`post_index` ambiguously means "after index"

weedcoco/index/indexing.py
```diff
@@ -0,0 +1,142 @@
...
+
+        files = [
+            file for file in os.listdir(self.temp_index_dir) if file.endswith(".json")
+        ]
+        for item in files:
+            requests.post(
```
jnothman, line 123: It's quite risky to ignore the response status code, for both the query in bulk and individual index entries. The Python elasticsearch package could be used to help manage error codes.

weedcoco/index/thumbnailing.py
```diff
@@ -0,0 +1,27 @@
+import argparse
+import os
+import pathlib
+from PIL import Image
+
+
+def thumbnailing(thumbnails_dir, repository_dir, THUMBNAIL_SIZE=(300, 300)):
+    for dirpath, dirnames, filenames in os.walk(repository_dir):
+        for filename in filenames:
+            if filename.lower().endswith((".png", ".jpg", ".jpeg", ".tiff")):
```
jnothman, line 10: should we only be processing files that do not already exist in the thumbnails dir?

jnothman (commented):
Thanks for your work on this. Many of my comments are stylistic, others are more about programming defensively.

I am concerned that at the moment most of the code is untested, which means that changes that introduce errors (even simple AttributeErrors) will not be immediately caught.
I am trying to think about what kinds of basic tests we could put in for indexing and thumbnailing without it being too burdensome.

For now I want to assume that the indexed data format is flexible, so I wouldn't want to test every aspect of it, but it would be good to at least run the code in test suites.

The existing indexing script, `search/scripts/weedcoco_to_elastic_index_bulk.sh` is retained in this PR, even though it is being superseded. We should plan to remove it.
The existing indexing script is currently tested by the [frontend workflow](https://github.com/Sydney-Informatics-Hub/Weed-ID-Interchange/blob/master/.github/workflows/frontend.yml) since it must succeed in order for [`indext_test_images.sh`](https://github.com/Sydney-Informatics-Hub/Weed-ID-Interchange/blob/d7ae6ba709e9652999000674f9ca3655c9ebb10a/.github/workflows/frontend.yml#L27) to complete with exit status 0. (It would be nice if we then checked that the right number of images had been indexed.)
We should have similar integration testing of the deposit, thumbnail and indexing logic in the frontend workflow. This will ensure that changes to the code that result in errors will more often than not be caught.

weedcoco/index/thumbnailing.py
```diff
@@ -0,0 +1,20 @@
...
+ap = argparse.ArgumentParser(description=__doc__)
+ap.add_argument("--thumbnails-dir", type=pathlib.Path, required=True)
+ap.add_argument("--repository-dir", type=pathlib.Path, required=True)
+args = ap.parse_args()
+
+THUMBNAIL_SIZE = (300, 300)
+
+for dirpath, dirnames, filenames in os.walk(args.repository_dir):
+    for filename in filenames:
+        if filename.lower().endswith((".png", ".jpg", ".jpeg", ".tiff")):
```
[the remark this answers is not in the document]
ElevnLi: > We should have similar integration testing of the deposit, thumbnail and indexing logic in the frontend workflow

Those tests can be integrated with pytest right?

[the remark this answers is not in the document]
jnothman: > Those tests can be integrated with pytest right?

Only if you write them to not require a live elasticsearch instance.

weedcoco/index/indexing.py
ElevnLi: I think for now when we only use the bulk post to interact with ElasticSearch and there doesn't show explicit shortage of either speed or capabilities, and the package is bit overkilling with extra parameters and uncertainty to care about. That package is more for enabling different kinds of customised search and XPack APIs which we most likely won't be making use of.
The attributes included in response object should be sufficient in terms of our requirements IMO.

ElevnLi:
I fixed items that have been marked as resolved. These have been tested manually but I didn't implement the pytest due to lack of ElasticSearch env. I'm not too sure about how to test it properly from frontend workflow with ElasticSearch environment, which we can discuss in Monday's catchup.

weedcoco/index/indexing.py
```diff
@@ -3,43 +3,52 @@
...
 import json
 import requests
-from math import ceil
+import urllib
 
 
 class ElasticSearchIndex:
+
+    """
+    Parameters:
```
jnothman, line 12: What docstring standard is this following? It's almost [Google Style](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html), but there they would have:
```suggestion
    Args:
```

And in weedcoco/importers/voc.py we have [one function](https://github.com/Sydney-Informatics-Hub/Weed-ID-Interchange/blob/d7ae6ba709e9652999000674f9ca3655c9ebb10a/weedcoco/importers/voc.py#L41-L49) with numpydoc-style docstrings. I'm happy to change it if you prefer Google style, but we should be consistent.

weedcoco/index/indexing.py
```diff
@@ -92,36 +101,39 @@ def _flatten(src, dst, prefix):
...
+        """
+        Send post request to ElasticSearch
+        """
+        for index_batch in self.generate_batches():
+            res = requests.post(
+                urllib.parse.urljoin(self.es_url, "_bulk"),
+                data=index_batch,
                 headers={"content-type": "application/json"},
             )
+            print(res.headers, res.status_code, res.content)
```
jnothman, line 136: Will this raise an error if one/all of the indexing commands fail?

Is there a reason not to use the Python elasticsearch library to do this, as well as the batching, for us?

weedcoco/index/thumbnailing.py
```diff
@@ -2,17 +2,20 @@
...
-            if filename.lower().endswith((".png", ".jpg", ".jpeg", ".tiff")):
-                image = Image.open(f"{dirpath}/{filename}")
-                image.thumbnail(THUMBNAIL_SIZE)
-                if filename[:2] not in os.listdir(thumbnails_dir):
-                    os.mkdir(thumbnails_dir / filename[:2])
-                image.save(thumbnails_dir / filename[:2] / filename)
+            if check_if_approved_image_format(filename):
+                image_path = thumbnails_dir / filename[:2] / filename
+                if not image_path.is_file():
+                    image = Image.open(f"{dirpath}/{filename}")
```
jnothman, line 14: (Maybe we should also check that `image.format` is one of the allowed types, as filename extension can be misleading.)

weedcoco/index/indexing.py
```diff
@@ -0,0 +1,149 @@
...
+                    )
+                yield blobs
+
+    def post_to_index(self):
+        """
+        Send post request to ElasticSearch
+        """
+        for index_batch in self.generate_batches():
+            helpers.bulk(self.es_client, index_batch)
+
```
ElevnLi, line 136: I'm not sure what exceptions need to be logged here.
Possible results are as https://elasticsearch-py.readthedocs.io/en/7.9.1/exceptions.html

weedcoco/index/indexing.py
```diff
@@ -0,0 +1,140 @@
...
+        if (
+            self.temp_index_dir == "temp"
+            or not os.path.isdir(self.temp_index_dir)
+            and "temp" not in os.listdir()
+        ):
+            self.temp_index_dir = "temp"
+            os.mkdir(self.temp_index_dir)
+
+    def modify_coco(self):
```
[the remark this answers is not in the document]
jnothman: I still think the name here isn't right but it's not essential for this PR.

jnothman (approved):
Let's merge this. Despite formally low code coverage, we've got good coverage if we include the frontend CI. (It would of course be nice to have pytest coverage for weedcoco.index too)

jnothman:
Thanks @ElevnLi

PR #195, opened 2021-02-01
add link to image card

ElevnLi:


jnothman:
Please state fixes #issuenumber in the PR description. I think this fixes two issues?

search/src/Components/wrapper/reactive_search.js
```diff
@@ -163,7 +163,7 @@ class ReactiveSearchComponent extends Component {
 							<ReactiveList.ResultCardsWrapper>
 								{
 									data.map(item => (
-										<ResultCard key={item._id}>
+										<ResultCard key={item._id} href={esURL + item.dataset_url}>
```
jnothman, line 166: I note that esURL is now a bad name. Maybe now is a good time to change it to rootURL?

jnothman: Are you assured that there is a `/` at the end of esURL?

Is it sufficient to just have the following which should mean the reference is relative to the domain root?
```suggestion
										<ResultCard key={item._id} href={"/" + item.dataset_url}>
```

ElevnLi:
fixes #169 
fixes #170

jnothman (commented):
I think the user may find it a bit surprising that clicking the ResultCard shows the entire dataset. I think it would be better to have a link within the card called "Dataset", perhaps with a tooltip that shows the dataset name, or with the name always displayed.

search/src/Components/wrapper/reactive_search.js
ElevnLi: this var name is everywhere for now, including other branches. Let's clean this together in another pr after merging together, maybe export it in App.js. using "/" perhaps will work but it needs to be tested on vm.

ElevnLi:
> I think the user may find it a bit surprising that clicking the ResultCard shows the entire dataset. I think it would be better to have a link within the card called "Dataset", perhaps with a tooltip that shows the dataset name, or with the name always displayed.

Can you give an example with an image maybe?

jnothman:
Simplest version:
image

i.e.

```html
<div><a title="Crop/Weed Field Image Dataset" href="datasets/">See Dataset</a></div></article>
```

A more sophisticated version would show a hovercard summarising the dataset.

jnothman:
Here's that code I drafted in ReactiveSearchComponent to get a mapping of upload ID to name. Not sure if it works!
```js
        const datasetNames = {};
        axios.get(baseURL + 'api/upload_list/').then(result => result.map((dataset) =>
            {datasetNames[dataset.upload_id] = dataset.name;}
        ));
```

search/src/Components/wrapper/reactive_search.js
```diff
@@ -179,6 +179,7 @@ class ReactiveSearchComponent extends Component {
 												</ul>
 												{" in " + item.agcontext__growth_stage_texts + " " + item.agcontext__crop_type}
 											</ResultCard.Description>
+											<div><a title={item.info ? item.info.name : ""} href={esURL + item.dataset_url}>See Dataset</a></div>
```
jnothman, line 182: this doesn't work yet, right?

ElevnLi: Indexing upload_id currently not working, coz previously was indexing dataset url. I'm troubleshooting now.

search/src/Components/wrapper/upload.js
```diff
@@ -20,13 +20,21 @@ const useStyles = (theme) => ({
 })
 
 const baseURL = new URL(window.location.origin);
+const upload_status_mapping = {
```
jnothman, line 23: This is unrelated to the present PR, isn't it? Can we pull it into a separate PR?

search/django/requirements.txt
```diff
@@ -6,4 +6,4 @@ celery==4.4.7
 django-celery-results
 django-celery-beat
 redis
-https://github.com/Sydney-Informatics-Hub/Weed-ID-Interchange/archive/ingestion.zip
\ No newline at end of file
+https://github.com/Sydney-Informatics-Hub/Weed-ID-Interchange/archive/image-link-dataset.zip
```
jnothman, line 9: This will need to become `master` upon merge, no?

jnothman (commented):
Excellent. In the future, we might consider a full hovercard for datasets.

search/django/requirements.txt
ElevnLi: after merging yes

PR #423, opened 2021-04-08
Store log

ElevnLi:


ElevnLi:
It seems the change on configs has a considerable lag. It started to work after not being touched for some time. It's using UTC time for now.
Screen Shot 2021-04-12 at 10 50 07 am

search/fluent.conf
```diff
@@ -0,0 +1,19 @@
...
+    time_key time
+    time_format %Y-%m-%dT%H:%M:%S
+    time_type string
+    localtime true
+  </parse>
+</source>
+
+<match docker.logs>
+  @type file
+  path /output/test.log
```
jnothman, line 18: Can we change this filename?

jnothman (commented):
It seems to be storing rolling logs, but would it not be normal for old logs to be compressed?

jnothman (approved):

PR #432, opened 2021-04-20
Improved error message reporting upon upload of images and invalid weedcoco

ElevnLi:


search/django/weedid/views.py
```diff
@@ -62,6 +62,16 @@ def upload(request):
         upload_dir, upload_id = setup_upload_dir(os.path.join(UPLOAD_DIR, str(user.id)))
         weedcoco_path = store_tmp_weedcoco(file_weedcoco, upload_dir)
         create_upload_entity(weedcoco_path, upload_id, user.id)
+    except JsonValidationError as e:
+        traceback.print_exc()
+        return HttpResponseForbidden(
```
jnothman, line 67: I'd prefer HTTP 400 than 403. 403 is quite specifically about preventing access to a resource. We will want this response to be a JSON blob.

weedcoco/validation.py
```diff
@@ -20,6 +19,18 @@ class ValidationError(Exception):
     pass
 
 
+class JsonValidationError(ValidationError):
+    def __init__(self, message, jsonschema_errors=None):
+        super().__init__(message)
+        self.jsonschema_errors = jsonschema_errors
+
+    def error_path_message(self):
```
jnothman, line 27: maybe call this `get_error_details`

weedcoco/validation.py
```diff
@@ -20,6 +19,18 @@ class ValidationError(Exception):
...
+        return [
+            (f"Error in {' / '.join(str(p) for p in error.path)}", error.message)
```
jnothman, line 29: Why build strings when we want this to be transmitted to the client side as a rich JSON blob in any case?

weedcoco/validation.py
```diff
@@ -29,20 +40,20 @@ def validate_json(weedcoco, schema="weedcoco", schema_dir=SCHEMA_DIR):
...
-    try:
-        jsonschema.validate(
-            weedcoco,
-            schema=main_schema,
-            resolver=jsonschema.RefResolver(schema_uri, main_schema, store=ref_store),
+    validator = Draft7Validator(main_schema)
+    validator.resolver = RefResolver(schema_uri, main_schema, store=ref_store)
+    errors = [error for error in validator.iter_errors(weedcoco)]
+    if len(errors):
+        raise JsonValidationError(
```
jnothman, line 53: Please add tests of this behaviour to `test_validate.py`.

jnothman (commented):
Thank you!

jnothman:
That last commit could have happily been a separate branch that could be merged without affecting this PR.

weedcoco/validation.py
ElevnLi: My thinking of design is that the error handling in the frontend only takes error message from BadRequest. If we want to send the wholes list of errors, might end up having an overweight exception discussion in the front.

jnothman: I don't think there's anything wrong with passing a kb of error messages back to the frontend, as long as the frontend limits/paginates the output. Or the backend can limit the total number of errors reported.

```
{
 "error_type": "jsonschema",
 "n_errors_found": "100",
 "error_details": [  # lists first 10 errors, say
    "path": ["categories", 0, "name"],
    "message": ...,
    "schema": ...,
 ]
}
```


Another thought: if the number of errors is really large, then it's likely that the number of _distinct_ errors is much smaller. As such you could instead group the content by message/schema:
```
{
 "error_type": "jsonschema",
 "n_errors_found": "100",
 "error_details": [
    "instances": [
      {"path": ["categories", 0, "name"], "value": "weed 1"},
      {"path": ["categories", 1, "name"], "value": "weed 2"},
      {"path": ["categories", 2, "name"], "value": "weed 3"}
    ],
    "message": ...,
    "schema": ...,
 ]
}
```

jnothman: Hmm... that doesn't work because the value is included in the error message.

But you could group failures by the `schema_path` or by the path without any array indices, like `["categories", "name"]`.

So the UI could say:
```
Total 11 errors

10 errors in the `name` field of categories

Showing first 3:
* 'weed 1' does not match ...
* 'weed 2' does not match ...
* 'weed 3' does not match ...

About the `name` field:
          A category for an annotation, consisting of an agricultural role and
          optionally a biological name, separated by ": ". Currently allowed
          agricultural roles are "weed" and "crop".  The "none" category may be
          used to indicate that no weed or crop is identified in the image.
          Annotations should be labelled with the most specific category possible,
          usually "weed: <species name>", but super categories may be more general.
          Biological names should be lowercase.
          For an unspecified species of weed, "weed: UNSPECIFIED" may be used.
          For example, to label an oat crop in an image, the category name would be
          "crop: avena sativa". Its super categories may include "crop: grasses"
          and "crop". Oats as a volunteer crop would be "weed: avena sativa".

1 error in the `bbch_growth_range` field of agcontexts:
* ....


About the `bbch_growth_range` field:
      BBCH Growth Stage range
      A pair of BBCH Growth Stage codes (00-99) representing the minimum and
      maximum crop growth stage in the images.
      If this AgContext is not in a cropping environment, or is in fallow, use value "na".
```

jnothman: But basically, for now, return all the errors in a structured form, and the client side can digest them for display, I think.

ElevnLi:
This returns back something like below:
```
{
    "error_type":"jsonschema",
    "n_errors_found":"3",
    "error_details":[
        {
            "path":[
                "categories",
                0,
                "name"
            ],
            "value":"weed: 1",
            "message":"'weed: 1' does not match '^((crop|weed): ([a-z][a-z ]+|UNSPECIFIED)|crop|weed|none)$'",
            "schema":{
                "type":"string",
                "pattern":"^((crop|weed): ([a-z][a-z ]+|UNSPECIFIED)|crop|weed|none)$",
                "description":"A category for an annotation, consisting of an agricultural role and\noptionally a biological name, separated by \": \". Currently allowed\nagricultural roles are \"weed\" and \"crop\".  The \"none\" category may be\nused to indicate that no weed or crop is identified in the image.\n\nAnnotations should be labelled with the most specific category possible,\nusually \"weed: <species name>\", but super categories may be more general.\nBiological names should be lowercase.\nFor an unspecified species of weed, \"weed: UNSPECIFIED\" may be used.\n\nFor example, to label an oat crop in an image, the category name would be\n\"crop: avena sativa\". Its super categories may include \"crop: grasses\"\nand \"crop\". Oats as a volunteer crop would be \"weed: avena sativa\"."
            }
        },
        {
            "path":[
                "categories",
                1,
                "name"
            ],
            "value":"weed: 2",
            "message":"'weed: 2' does not match '^((crop|weed): ([a-z][a-z ]+|UNSPECIFIED)|crop|weed|none)$'",
            "schema":{
                "type":"string",
                "pattern":"^((crop|weed): ([a-z][a-z ]+|UNSPECIFIED)|crop|weed|none)$",
                "description":"A category for an annotation, consisting of an agricultural role and\noptionally a biological name, separated by \": \". Currently allowed\nagricultural roles are \"weed\" and \"crop\".  The \"none\" category may be\nused to indicate that no weed or crop is identified in the image.\n\nAnnotations should be labelled with the most specific category possible,\nusually \"weed: <species name>\", but super categories may be more general.\nBiological names should be lowercase.\nFor an unspecified species of weed, \"weed: UNSPECIFIED\" may be used.\n\nFor example, to label an oat crop in an image, the category name would be\n\"crop: avena sativa\". Its super categories may include \"crop: grasses\"\nand \"crop\". Oats as a volunteer crop would be \"weed: avena sativa\"."
            }
        },
        {
            "path":[
                "agcontexts",
                0,
                "bbch_growth_range"
            ],
            "value":{
                "min":-1,
                "max":24
            },
            "message":"{'min': -1, 'max': 24} is not valid under any of the given schemas",
            "schema":{
                "oneOf":[
                    {
                        "type":"object",
                        "title":"BBCH Range",
                        "required":[
                            "min",
                            "max"
                        ],
                        "additionalProperties":false,
                        "properties":{
                            "min":{
                                "type":"integer",
                                "minimum":0,
                                "maximum":99,
                                "description":"Minimum BBCH growth stage (00-99) in the dataset"
                            },
                            "max":{
                                "type":"integer",
                                "minimum":0,
                                "maximum":99,
                                "description":"Maximum BBCH growth stage (00-99) in the dataset"
                            }
                        }
                    },
                    {
                        "type":"string",
                        "title":"Not Applicable",
                        "const":"na"
                    }
                ],
                "description":"BBCH Growth Stage range\nA pair of BBCH Growth Stage codes (00-99) representing the minimum and maximum crop growth stage in the images.\nIf this AgContext is not in a cropping environment, or is in fallow, use value \"na\"."
            }
        }
    ]
}
```

ElevnLi:
Screen Shot 2021-04-21 at 3 36 32 pm

search/src/Components/error/utils.js
```diff
@@ -0,0 +1,17 @@
...
+export const jsonSchemaTransform = (details) => {
+    const transform = {}
+    const pathKey = (path) => `'${path[2]}' field of ${path[0]}`
+    for (const error_detail of details.error_details) {
+        let path = pathKey(error_detail.path)
+        if (path in transform) {
+            transform[path].instances.push({path: error_detail.path.join(" / "), message: error_detail.message})
+        } else {
+            transform[path] = {description: error_detail.schema.description, instances: [{path: error_detail.path.join(" / "), message: error_detail.message}]}
+        }
```
jnothman, line 14: This is much easier to maintain:
```suggestion
        if (!(path in transform)) {
            transform[path] = {description: error_detail.schema.description, instances: []}
        }
        transform[path].instances.push({path: error_detail.path.join(" / "), message: error_detail.message});
```

search/src/Components/error/utils.js
```diff
@@ -0,0 +1,17 @@
+export const jsonSchemaTitle = (res) => {
+    return `Total ${res.n_errors_found} ${res.n_errors_found > 1? "errors": "error"}`
+}
+
+export const jsonSchemaTransform = (details) => {
+    const transform = {}
+    const pathKey = (path) => `'${path[2]}' field of ${path[0]}`
```
jnothman, line 7: This won't behave appropriately if:
* the error is in `info`, since it is not an array
* the bad field is deeper, such as `annotations.0.segmentation.counts`

So you might want to do something like:
```js
const namedPaths = path.filter(word => typeof word == typeof '');
```
For pathKey you might want to leave it as `namedPaths.join('.')` but for display you might want to switch based on path length such that:
* ['categories'] -> categories
* ['categories', 'name'] -> name field of categories
* ['annotations', 'segmentation', 'counts'] -> counts field of segmentation in annotations

or don't bother with that level of intricacy initially and just display namedPaths.join('.') when it's not of length 2.

I'd probably not transform for the purpose of display until as late as possible, since you might want to format those path elements. Constructing a string here means that you can't format it as react components later.

Deserves unit tests.

search/src/Components/error/utils.js
```diff
@@ -0,0 +1,17 @@
...
+    for (const error_detail of details.error_details) {
+        let path = pathKey(error_detail.path)
+        if (path in transform) {
+            transform[path].instances.push({path: error_detail.path.join(" / "), message: error_detail.message})
```
jnothman, line 11: Joining on `.` might be more traditional for json?

search/src/Components/error/details.js
```diff
@@ -0,0 +1,37 @@
...
+import {jsonSchemaTransform} from './utils';
+
+
+const JsonSchemaDetails = (props) => {
+    const {details} = props
+    const tfError = jsonSchemaTransform(details)
+    const tfDisplay = Object.keys(tfError).map((path) => {
+        return (
+            <div>
+                <p>{tfError[path].instances.length} {tfError[path].instances.length > 1 ? "errors": "error"} in the {path}</p>
```
jnothman, line 11: I'd be tempted to put these all in accordions. WDYT? I think the different errors need more visual distinction, certainly, either using material Paper for each, or at least making this heading size.

Idiomatic react would use Typography here, rather than p.

search/src/Components/error/details.js
```diff
@@ -0,0 +1,37 @@
...
+                {tfError[path].instances.map((error) => <p>{error.message} @ {error.path}</p>)}
```
jnothman, line 12:
```suggestion
                <ol>
                {tfError[path].instances.map(error =>
                  <li key={error.path}>In {error.path}: {error.message}</li>
                )}
                </ol>
```
There might be a material-style component better than `<li>`

search/src/Components/error/details.js
```diff
@@ -0,0 +1,37 @@
...
+                <p>Description:</p>
```
jnothman, line 13:
```suggestion
                <p>Description of this field:</p>
```

search/src/Components/error/details.js
```diff
@@ -0,0 +1,37 @@
...
+                <p>{tfError[path].description}</p>
```
jnothman, line 14:
```suggestion
                <ReactMarkdown source={tfError[path].description} />
```

jnothman (commented):
Looking like a nice first draft

Please also test "required field missing" cases.

weedcoco/tests/testcases.py
```diff
@@ -0,0 +1,805 @@
...
+            {
+                "path": ["categories", 0, "name"],
+                "value": "foobar",
+                "message": "'foobar' does not match '^((crop|weed): ([a-z][a-z ]+|UNSPECIFIED)|crop|weed|none)$'",
+                "schema": {
+                    "type": "string",
+                    "pattern": "^((crop|weed): ([a-z][a-z ]+|UNSPECIFIED)|crop|weed|none)$",
+                    "description": 'A category for an annotation, consisting of an agricultural role and\noptionally a biological name, separated by ": ". Currently allowed\nagricultural roles are "weed" and "crop".  The "none" category may be\nused to indicate that no weed or crop is identified in the image.\n\nAnnotations should be labelled with the most specific category possible,\nusually "weed: <species name>", but super categories may be more general.\nBiological names should be lowercase.\nFor an unspecified species of weed, "weed: UNSPECIFIED" may be used.\n\nFor example, to label an oat crop in an image, the category name would be\n"crop: avena sativa". Its super categories may include "crop: grasses"\nand "crop". Oats as a volunteer crop would be "weed: avena sativa".',
```
jnothman, line 143: I think it's a bad idea to reproduce the schema here, since small changes to the schema will need to be replicated in the test suite. They're not essential to testing our code either.

I don't have a simple solution except to break up the assertion... I struggled with this internally, so I've had a hack at how I might write that:

```py
def check_error_details(actual, expected):
    def _everything_but_details(obj):
        return ({k: v for k, v in obj.items() if k != "error_details"}

    def _stable_details(obj):
        return ({k: v for k, v in obj["error_details"].items() if k not in ["schema", "message"]}

    assert actual.keys() == expected.keys()
    assert _everything_but_details(actual) == _everything_but_details(expected)
    assert actual["error_details"].keys() == expected["error_details"].keys()
    assert _stable_details(actual) == _stable_details(expected)
```

ElevnLi: I think it's easier just to remove the schema from actual, as the structure of it can be predicted

ElevnLi:
Screen Shot 2021-04-23 at 1 02 08 pm

ElevnLi:
Screen Shot 2021-04-27 at 2 22 54 pm
Screen Shot 2021-04-27 at 2 23 04 pm
It won't allow users to submit before receiving all required images mentioned in the json.

search/src/Components/error/details.js
```diff
@@ -0,0 +1,75 @@
...
+const JsonSchemaDetails = (props) => {
+    const {details} = props
+    const classes = useStyles();
+    const tfError = jsonSchemaTransform(details)
+    const tfDisplay = Object.keys(tfError).map((path) => {
+        return (
+            <div className={classes.root}>
+                <Accordion>
+                    <AccordionSummary expandIcon={<ExpandMoreIcon />}>
+                        <Typography className={classes.heading}>{tfError[path].instances.length} {tfError[path].instances.length > 1 ? "errors": "error"} in the {path}
```
jnothman, line 32:
```suggestion
                        <Typography className={classes.heading}>
                            {tfError[path].instances.length} {tfError[path].instances.length > 1 ? "errors": "error"} in the {path}
```

search/src/Components/upload/react-images-upload/index.js
```diff
@@ -129,11 +130,25 @@ class ReactImageUploadComponent extends React.Component {
...
 
+  syncImageErrorMessage(updatedFilesName) {
+    const missingImagesAmount = this.props.images.length - updatedFilesName.length;
+    const missingImages = [...this.props.images].filter(image => !updatedFilesName.includes(image));
+    if (missingImagesAmount == 0 && new Set(missingImages).size === 0) {
+        this.props.handleImageReady(true);
+        this.props.handleErrorMessage("");
+    } else {
+        this.props.handleImageReady(false);
+        this.props.handleErrorMessage(`${missingImagesAmount} ${missingImagesAmount > 1 ? "images" : "image"} missing`, {error_type: "image", missingImages: missingImages});
```
jnothman, line 148: I'm not entirely with the separation of concerns here... It feels like you're pushing too much weed-ai-specific specific work onto the upload component. That is, I think it's good if `ReactImageUploadComponent` calls a handler, e.g. `handleUploaded(updatedFiles)` or `handleUploaded(updatedFileNames)`, but the translation of that to an error message should be weed-ai logic, not in react-images-upload code.

What other changes had you made to react-images-upload?

search/src/Components/error/dialog.js
```diff
@@ -0,0 +1,73 @@
...
+
+  const handleClickOpen = () => {
+    setOpen(true);
+  };
+  const handleClose = () => {
+    setOpen(false);
+  };
+
+  return (
+    <div>
```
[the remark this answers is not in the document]
ElevnLi: react function can only return a single jsx element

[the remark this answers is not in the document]
jnothman: you can't return a React.Fragment? or an array of elements?

search/src/Components/upload/react-images-upload/index.js
ElevnLi: You mean move syncImageErrorMessage outside and rename it to handleUploaded?

jnothman: Effectively, yes, but the goal is that ReactImageUploadComponent doesn't deal with things that are not about the image uploading, i.e. it shouldn't care about which files are expected/unexpected.

jnothman (commented):
Thanks.

jnothman (approved):

PR #443, opened 2021-05-04
Category mapper

ElevnLi:


ElevnLi:
Screen Shot 2021-05-04 at 4 48 49 pm

ElevnLi:
Screen Shot 2021-05-04 at 4 48 33 pm

weedcoco/schema/compatible-coco.yaml
```diff
@@ -18,4 +18,4 @@ properties:
   categories:
     type: array
     items:
-      $ref: https://weedid.sydney.edu.au/schema/Category.json#/
```
jnothman, line 21: The thing that defined "compatible-coco" was that it had the weedcoco category requirements. If we're getting rid of the category requirements, we should call this format "coco", not "compatible-coco".

search/django/weedid/utils.py
```diff
@@ -31,6 +32,34 @@ def setup_upload_dir(upload_userid_dir):
     return upload_dir, upload_id
 
 
+def set_categories(weedcoco_path, categories):
+    with open(weedcoco_path, "r") as jsonFile:
+        data = json.load(jsonFile)
+    new_categories = []
+    for category in categories:
+        if category["scitific_name"]:
```
jnothman, line 40: Please correct "scitific" to "scientific" throughout

search/django/weedid/utils.py
```diff
@@ -118,3 +147,20 @@ def retrieve_listing_info(query_entity, awaiting_review):
...
         "contributor_email": query_entity.user.email if awaiting_review else "",
     }
+
+
+def retrieve_category_name(category):
+    if re.fullmatch(r"(crop|weed): ([a-z][a-z ]+|UNSPECIFIED)", category["name"]):
+        return {
+            "id": category["id"],
+            "category": category["name"].split(": ")[0],
+            "name": category["name"].split(": ")[1],
```
jnothman, line 157: shouldn't this be `category["name"]`?

search/django/weedid/utils.py
```diff
@@ -118,3 +147,20 @@ def retrieve_listing_info(query_entity, awaiting_review):
         "contributor": query_entity.user.username,
         "contributor_email": query_entity.user.email if awaiting_review else "",
     }
+
+
+def retrieve_category_name(category):
+    if re.fullmatch(r"(crop|weed): ([a-z][a-z ]+|UNSPECIFIED)", category["name"]):
```
jnothman, line 153: A scientific name can have non-alphabetic characters in it, notably characters like `ö`. I think it's fine to test if it begins `crop` or `weed`.

We also allow the category name to be "other".

search/django/weedid/utils.py
```diff
@@ -118,3 +147,20 @@ def retrieve_listing_info(query_entity, awaiting_review):
...
+        return {
+            "id": category["id"],
+            "category": category["name"].split(": ")[0],
```
jnothman, line 156: let's call this `role`, rather than overload "category".

search/src/Components/upload/uploader_category_mapper.js
```diff
@@ -0,0 +1,111 @@
...
+                                    <FormControl className={classes.formControl}>
+                                        <InputLabel id="category">Category</InputLabel>
+                                        <Select
+                                        labelId="category"
+                                        id="category"
+                                        value={category.category}
+                                        onChange={e => {this.changeCategory(e, index)}}
+                                        >
+                                            <MenuItem value={"crop"}>crop</MenuItem>
+                                            <MenuItem value={"weed"}>weed</MenuItem>
```
jnothman, line 82: we might need to allow for "remove category" as well, perhaps not for now.

search/src/Components/upload/uploader_category_mapper.js
```diff
@@ -0,0 +1,111 @@
...
+    row: {
+        display: 'flex',
+        alignItems: 'baseline'
+    },
+    formControl: {
+        margin: theme.spacing(1),
+        minWidth: 120,
+    },
+    span: {
+        marginRight: '1em'
```
jnothman, line 25: I think minWidth might better serve the purpose... But so would a table or another grid-based layout.

search/src/Components/upload/uploader_category_mapper.js
```diff
@@ -0,0 +1,111 @@
...
+    },
+    row: {
+        display: 'flex',
+        alignItems: 'baseline'
+    },
+    formControl: {
+        margin: theme.spacing(1),
+        minWidth: 120,
+    },
+    span: {
```
jnothman, line 24: span is a bit of an unhelpful name

search/src/Components/upload/uploader_category_mapper.js
```diff
@@ -0,0 +1,111 @@
...
+                        this.state.categories.map((category, index) => {
+                            return (
+                                <li className={classes.row}>
+                                    <p className={classes.span}>{category.name} is a</p>
+                                    <FormControl className={classes.formControl}>
+                                        <InputLabel id="category">Category</InputLabel>
+                                        <Select
+                                        labelId="category"
+                                        id="category"
+                                        value={category.category}
```
jnothman, line 78: Let's make weed the default?
```suggestion
                                        value={category.category || "weed"}
```

search/src/Components/upload/uploader_category_mapper.js
```diff
@@ -0,0 +1,111 @@
...
+        this.modifyCategories(index, "scitific_name", e.target.value);
+        this.props.handleCategoryStatus(false);
+        this.props.handleErrorMessage("init");
+    }
+
+    render() {
+        const { classes } = this.props;
+        return (
+            <React.Fragment>
+                <p className={classes.summary}>{this.state.categories.length} {this.state.categories.length > 1 ? 'categories' : 'category'} need mapping to weedcoco categories</p>
```
jnothman, line 66: This number should be updated as form values are filled in, such that if the category names are already valid weedcoco categories, this may start with "0 categories need mapping".

The 0 case should have plural "categories":
```suggestion
                <p className={classes.summary}>{this.state.categories.length} {this.state.categories.length != 1 ? 'categories' : 'category'} need mapping to weedcoco categories</p>
```

search/src/Components/upload/uploader_category_mapper.js
```diff
@@ -0,0 +1,111 @@
...
+                                </li>
+                            )
+                        })
+                    }
+                </ol>
+                <Button variant="contained"
+                        color="primary"
+                        className={classes.applyButton}
+                        onClick={() => {
+                            if (this.state.categories.length > 0 && this.state.categories.length === this.state.categories.filter(category => category.category.length > 0).length) {
```
jnothman, line 96: I think this is incorrectly checking `category.category.length`. Shouldn't it be category.scitific_name.length?

jnothman (commented):
I think that without some amount of automated testing of these components, you're likely to spend a lot of time unnecessarily performing manual testing.

weedcoco/schema/compatible-coco.yaml
ElevnLi: Do we want to keep both coco and compatible-coco? or just rename it to coco instead.

search/django/weedid/utils.py
ElevnLi: I'm differentiating 2 scenarios that when the name already matches with the pattern `r"(crop|weed): ([a-z][a-z ]+|UNSPECIFIED)"`, I extract category and name from it and give the frontend a preset of the current category but also let them change it.

search/django/weedid/utils.py
ElevnLi: Should I add "other" to the category schema? It's currently missing.

search/src/Components/upload/uploader_category_mapper.js
ElevnLi: Is the scientific_name mandatory? I was checking if every entry has a category.

weedcoco/schema/Category.yaml
```diff
@@ -6,7 +6,7 @@ allOf:
     properties:
       name:
         type: string
-        pattern: "^((crop|weed): ([a-z][a-z ]+|UNSPECIFIED)|crop|weed|none)$"
+        pattern: "^((crop|weed|other): ([a-z][a-z ]+|UNSPECIFIED)|crop|weed|none)$"
```
jnothman, line 9: "other" doesn't need a species, so I guess it should be
```suggestion
        pattern: "^((crop|weed): ([a-z][a-z ]+|UNSPECIFIED)|crop|weed|other|none)$"
```

but then "other" needs a description below.

Maybe "other" is stupid and we should leave it out. At one point I thought we'd use it to represent other objects like a boot or a piece of machinery captured in the image. But maybe it's not useful. As they say, YAGNI.

jnothman:
Part of the response re continuing to support compatible-coco is that it would be helpful to validate against compatible-coco when Apply is clicked, once species validation (#420) is enabled

weedcoco/schema/coco.yaml
```diff
@@ -0,0 +1,21 @@
+$schema: http://json-schema.org/draft-07/schema#
+$id: https://weedid.sydney.edu.au/schema/compatible-coco.json
```
jnothman, line 2:
```suggestion
$id: https://weedid.sydney.edu.au/schema/coco.json
```

weedcoco/schema/coco.yaml
```diff
@@ -0,0 +1,21 @@
...
+description: Minimal MS COCO format with WeedCOCO category naming scheme.
```
jnothman, line 3:
```suggestion
description: MS COCO format
```

search/src/Components/upload/uploader_category_mapper.js
```diff
@@ -0,0 +1,112 @@
...
+        this.modifyCategories(index, "scientific_name", e.target.value);
+        this.props.handleCategoryStatus(false);
+        this.props.handleErrorMessage("init");
+    }
+
+    render() {
+        const { classes } = this.props;
+        return (
+            <React.Fragment>
+                <p className={classes.summary}>{this.state.categories.length} {this.state.categories.length != 1 ? 'categories' : 'category'} need mapping to weedcoco categories</p>
```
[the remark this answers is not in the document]
ElevnLi: This idea doesn't look good to me. Firstly, how do we define a complete row? For those user already uploaded with both role and sci_name or for those they have clicked the apply button for once? then they cannot change their mind after applying it?
The case is that there won't be say 100 categories to fill in for each upload. I reckon just let user have an overview what they will upload. Every time they change the form, it will be required to click the Apply button again to sync it and enable Next button. Otherwise it's messy to just manage the state of categories.
The best way is to give a checkbox of only showing not-yet-applied categories, and that's a bit over engineering if there will be just a few.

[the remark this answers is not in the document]
jnothman: Above I defined nComplete... just the way you'd defined it in the onclick handler: https://github.com/Sydney-Informatics-Hub/Weed-AI/pull/443/files#r629820294

I should be clear: I don't think we need to make this minor UI change to colour the rows when they're incomplete. It was just to communicate how I would see a more polished version of this.

A more polished version would also have a different apply/next interaction, but what you've got here is fine.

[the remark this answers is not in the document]
jnothman: I think reducing the number of errors as the user fills rows in is good, or at worst harmless.

search/src/Components/upload/uploader_category_mapper.js
```diff
@@ -0,0 +1,112 @@
...
+                                this.props.handleCategories(this.state.categories)
+                                this.props.handleErrorMessage("")
+                                this.props.handleCategoryStatus(true)
+                            }
+                            else {
+                                console.log(this.state.categories)
+                                this.props.handleErrorMessage("Category or Scientific name missing")
+                                this.props.handleCategoryStatus(false)
+                            }
+                        }}>Apply</Button>
```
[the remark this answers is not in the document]
ElevnLi: The Next button currently is disabled until the user has successfully applied their changes. What if the default file that user uploaded already matched the rule. I think it's better to have Apply enabled and let Next do the constraint work.

ElevnLi:
Screen Shot 2021-05-11 at 3 33 46 pm

search/src/Components/upload/upload_stepper.js
```diff
@@ -214,10 +247,16 @@ class UploadStepper extends React.Component {
         })
     }
 
+    handleCategoryStatus(category_status){
+        this.setState({category_status: category_status})
```
jnothman, line 251: We should probably call this state `categoriesValid` or `categoriesComplete`. Status is not usually a bool, and its interpretation is unclear.

search/django/weedid/utils.py
```diff
@@ -118,3 +139,20 @@ def retrieve_listing_info(query_entity, awaiting_review):
         "contributor": query_entity.user.username,
         "contributor_email": query_entity.user.email if awaiting_review else "",
     }
+
+
+def parse_category_name(category):
```
jnothman, line 144: [Something like this might better belong in the weedcoco library, but it's okay here for now.]

jnothman (approved):
Apart from the naming of category_status, LGTM. To avoid more rounds, I'll merge and fix it up.

search/src/Components/upload/upload_stepper.js
jnothman: categoriesSaved might be better.

jnothman: because status is only true after Apply, I think

PR #447, opened 2021-05-07
E2e test with cypress

ElevnLi:


ElevnLi:
Screen Shot 2021-05-07 at 11 47 56 am

ElevnLi:
This is a version that works for login and upload weedcoco, but I reckon it's not the most elegant way doing it. I tried a bit from https://github.com/testing-library/cypress-testing-library/blob/97939da7d4707a71049884c0324c0eda56e26fc2/cypress/integration/find.spec.js but didn't make it to locate element easily like finding the sign in button even.

search/src/Components/auth/auth_prompt.js
```diff
@@ -49,11 +49,11 @@ export default function AuthPrompt(props) {
     <div>
       <div style={{display: 'flex', flexDirection: 'row', alignItems: 'center'}}>
         Please&nbsp;
-        <Button variant="outlined" color="primary" onClick={handleLogin}>
+        <Button id="sign_in_button" variant="outlined" color="primary" onClick={handleLogin}>
```
jnothman, line 52: Yeah, so this is definitely the kind of thing where the preferred testing code should _not_ use the ID. Rather, you should use use `cy.findByRole('button', {name: /Sign In/})`.

See https://testing-library.com/docs/queries/byrole/ which mentions the default `role` of buttons and the "accessible name" being the text of the button. See also "Rules of ARIA attribute usage by HTML element" at [the aria spec](https://www.w3.org/TR/html-aria).

ElevnLi: It worths trying out for different combination to actually use the api
Screen Shot 2021-05-07 at 1 06 43 pm

ElevnLi: Screen Shot 2021-05-07 at 1 10 58 pm

jnothman: Strange that there are multiple elements but we can't see them listed...?

jnothman:
There also seems to be a dep version conflict between `node` and `cypress`.

jnothman (changes requested):
Can we please also add the cypress test run to `.github/workflows/frontend.yml`?

jnothman:
This is now complaining that 001_image.png is missing. Did you forget to commit it? (perhaps use a low-quality jpeg to save space??)

jnothman:
I should clarify: it turns out that the `celery` and `django` jobs failed, since `.env` didn't exist. This is now rectified.

#457 improves handling such errors so that they lead to CI failure.
afc40d98 halved the runtime of the frontend.yml check by caching layers in the docker build.

jnothman:
It passes 🥂🥳🎉

jnothman:
Okay to take this out of Draft mode?

search/cypress/integration/weedai_upload.js
```diff
@@ -0,0 +1,38 @@
+describe('Test the overallworkflow', () => {
```
jnothman, line 1:
```suggestion
describe('Test the overall upload workflow', () => {
```

jnothman (approved):
lgtm

jnothman (approved):
Awesome refactor, thanks

PR #459, opened 2021-05-18
Voc uploader first draft

ElevnLi:


ElevnLi:
I'll just put my notes of findings here.
- metadata json form not working(stand alone and integrated)
- the way to save temporary voc files, having a random folder
- validation of xml files, just format?
- voc_to_coco, validate coco (no agcontexts), but create_upload_entity → make_upload_entity_fields → Weedcocostats, needs agcontexts in annotations

search/django/weedid/views.py
```diff
@@ -91,6 +95,74 @@ def upload(request):
...
+    user = request.user
+    if not (user and user.is_authenticated):
+        return HttpResponseForbidden("You dont have access to proceed")
+    try:
+        voc_dir = request.POST["voc_dir"]
+        voc = request.FILES["voc"]
+        upload_dir = os.path.join(UPLOAD_DIR, str(user.id), voc_dir)
+        store_tmp_voc(voc, upload_dir)
+    except Exception as e:
+        return HttpResponseBadRequest(str(e))
```
jnothman, line 110: I think we need to log the exception (with `print_exc`) as well as returning a meaningful message like this. I think this should also be done with a decorator. Django Rest Framework would do this kind of stuff for us, if we had adopted it.

search/django/weedid/views.py
```diff
@@ -91,6 +95,74 @@ def upload(request):
...
+    user = request.user
+    if not (user and user.is_authenticated):
+        return HttpResponseForbidden("You dont have access to proceed")
+    voc_dir = request.POST["voc_dir"]
+    voc_name = request.POST["voc_name"]
+    voc_to_remove = os.path.join(UPLOAD_DIR, str(user.id), voc_dir, voc_name)
+    if os.path.exists(voc_to_remove):
+        os.remove(voc_to_remove)
+        return HttpResponse(f"Removed {voc_name}")
+    else:
```
jnothman, line 127: I'd do this with try-except. Apart from anything else, it avoids race conditions. (This code could raise an exception if two identical remove_voc requests were issued at the same time.)

search/src/Components/upload/uploader_multiple.js
```diff
@@ -1,52 +1,84 @@
...
+      else if (status === 'error_file_size') {
+          props.handleErrorMessage("The file size exceeds the limitation")
+      }
+      else if (status === 'removed') {
+          removeVoc(file)
+      }
     }
 
     return (
       <Dropzone
```
jnothman, line 76: can we limit the file size to something small?

jnothman (commented):
The current temporary dir solution is okay, but it would be best if it was localised to a specific user ID so that one user could not modify another user's data, or create arbitrary directories in the MEDIA_ROOT. I'd be happier, I think, with not creating another temporary directory, but using the asynchronously generated `upload_id`, which can be validated... and we can expire and remove all the content together.

jnothman:
Re validation of XML files: if one file is bad it's going to be hard to handle. We'll have to have an all-or-nothing validation for now, which `voc_to_coco` will suffice for. I don't think we can make the errors more helpful for now.

search/src/Components/upload/uploader_multiple.js
```diff
@@ -1,52 +1,84 @@
...
-}
-
+const tmp_voc = Math.random().toString(36).slice(-8);
 const UploaderMultiple = (props) => {
     const baseURL = new URL(window.location.origin);
-    const getUploadParams = ({ file, meta }) => {
+
+    const removeVoc = file => {
         const body = new FormData()
-        body.append('upload_image', file)
```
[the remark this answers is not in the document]
ElevnLi: yeah this multiple uploader was replaced by another image uploader before.

search/django/weedid/utils.py
```diff
@@ -32,6 +32,24 @@ def store_tmp_weedcoco(weedcoco, upload_dir):
     return weedcoco_path
 
 
+def store_tmp_voc(voc, voc_dir):
+    fs = FileSystemStorage()
+    fs.save(os.path.join(voc_dir, voc.name), voc)
+
+
+def store_tmp_voc_coco(weedcoco, upload_dir):
+    if "agcontexts" not in weedcoco:
```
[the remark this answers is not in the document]
ElevnLi: makes sense, maybe call it when marking it as awaiting_review

search/django/weedid/views.py
```diff
@@ -91,6 +95,74 @@ def upload(request):
         )
 
 
+def upload_voc(request):
+    if not request.method == "POST":
```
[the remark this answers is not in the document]
ElevnLi: In another pr maybe

search/django/weedid/utils.py
```diff
@@ -32,6 +32,24 @@ def store_tmp_weedcoco(weedcoco, upload_dir):
     return weedcoco_path
 
 
+def store_tmp_voc(voc, voc_dir):
+    fs = FileSystemStorage()
+    fs.save(os.path.join(voc_dir, voc.name), voc)
```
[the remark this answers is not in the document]
ElevnLi: The api to receive, remove and submit voc is requesting id directly from the user in the request, so user cannot say, request a process of other users' voc even when he somehow knows the voc_id (generated randomly every time the uploader dialogue is opened) and user_id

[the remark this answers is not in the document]
jnothman: but in the current design we have no way to even trace the directory back to a particular user, which makes me uncomfortable.

search/django/weedid/views.py
jnothman: Please log the exception

search/django/weedid/views.py
```diff
@@ -122,8 +197,8 @@ def update_categories(request):
         validate(updated_weedcoco_json, schema="compatible-coco")
     except JsonValidationError as e:
         return HttpResponseBadRequest(e.message)
-    except Exception:
-        return HttpResponseNotAllowed("Failed to update categories")
+    except Exception as e:
+        return HttpResponseBadRequest(str(e))
```
jnothman, line 201: please log the exception with print_exc (or just leave the exception unhandled and let it return 50x)

search/src/Components/upload/uploader_voc.js
```diff
@@ -0,0 +1,93 @@
+import React from 'react';
+import 'react-dropzone-uploader/dist/styles.css';
+import Dropzone from 'react-dropzone-uploader';
+import {jsonSchemaTitle} from '../error/utils';
+import axios from 'axios';
+import Cookies from 'js-cookie';
+
+
+const tmp_voc = Math.random().toString(36).slice(-8);
```
jnothman, line 9: This is a global constant, and could be the same for every upload if the user does multiple in one session. Maybe we should set this inside the upload stepper?

jnothman (commented):
This otherwise appears good

search/src/Components/upload/uploader_voc.js
ElevnLi: This works as expected. If you put it inside, it applies to single selection of files, say if you upload one voc and decide to upload another one, they will go to different folder. Not sure why the js scope is working like this, but I came up with this solution by practice.

search/django/weedid/views.py
```diff
@@ -75,8 +79,80 @@ def upload(request):
...
+            Path(os.path.join(UPLOAD_DIR, str(user.id), voc_dir))
+        )
+        validate(weedcoco_json, schema="coco")
+        for image_reference in weedcoco_json["images"]:
+            images.append(image_reference["file_name"].split("/")[-1])
+        categories = [
+            parse_category_name(category) for category in weedcoco_json["categories"]
+        ]
+        upload_dir, upload_id = setup_upload_dir(os.path.join(UPLOAD_DIR, str(user.id)))
+        store_tmp_voc_coco(weedcoco_json, upload_dir)
```
jnothman, line 154: We should either move or delete the VOC files.

search/django/weedid/views.py
```diff
@@ -75,8 +79,80 @@ def upload(request):
...
+    if not request.method == "POST":
+        return HttpResponseNotAllowed(request.method)
+    user = request.user
+    if not (user and user.is_authenticated):
+        return HttpResponseForbidden("You dont have access to proceed")
+    try:
+        voc_dir = request.POST["voc_dir"]
+        voc = request.FILES["voc"]
+        upload_dir = os.path.join(UPLOAD_DIR, str(user.id), voc_dir)
+        store_tmp_voc(voc, upload_dir)
```
jnothman, line 108: Can we reject files > 1kb?

search/src/Components/upload/uploader_voc.js
jnothman: > If you put it inside, it applies to single selection of files

That's why I suggested initialising this as part of the stepper, not part of the hook. It's a unique token per upload, so that's context that can be maintained by the stepper.

ElevnLi: I put it in the uploader state now

search/django/weedid/views.py
ElevnLi: I set moving voc to upload_id/voc after user clicking the NEXT button, as after clicking submit button, user may still want to remove or modify voc. If the moving or deleting happen there, it will cause an error saying voc doesn't exist.

search/django/weedid/views.py
```diff
@@ -91,6 +102,97 @@ def upload(request):
...
 
 
+def upload_voc(request):
+    if not request.method == "POST":
+        return HttpResponseNotAllowed(request.method)
+    user = request.user
+    if not (user and user.is_authenticated):
+        return HttpResponseForbidden("You dont have access to proceed")
+    try:
+        voc_id = request.POST["voc_id"]
```
jnothman, line 112: please assert that voc_id does not contain `/`

search/django/weedid/views.py
```diff
@@ -91,6 +102,97 @@ def upload(request):
...
+        return HttpResponse(f"Uploaded {voc.name} to {upload_dir}")
+
+
+def remove_voc(request):
+    if not request.method == "POST":
+        return HttpResponseNotAllowed(request.method)
+    user = request.user
+    if not (user and user.is_authenticated):
+        return HttpResponseForbidden("You dont have access to proceed")
+    voc_id = request.POST["voc_id"]
```
jnothman, line 131: please assert that voc_id does not contain `/`

search/django/weedid/views.py
```diff
@@ -91,6 +102,97 @@ def upload(request):
...
+        return HttpResponseBadRequest(f"Error when removing {voc_name}")
+
+
+def submit_voc(request):
+    if not request.method == "POST":
+        return HttpResponseNotAllowed(request.method)
+    user = request.user
+    if not (user and user.is_authenticated):
+        return HttpResponseForbidden("You dont have access to proceed")
+    voc_id = request.POST["voc_id"]
```
jnothman, line 150: please assert that voc_id does not contain `/`

jnothman (approved):

PR #491, opened 2021-06-15
Improve upload experience for large numbers of files

ElevnLi:


search/src/Components/upload/react-images-upload/index.js
```diff
@@ -63,6 +64,8 @@ class ReactImageUploadComponent extends React.Component {
   onDropFile(e) {
     const files = e.target.files;
     const fileErrors = [];
+    const uploadedFiles = this.state.files.slice() ? this.state.files.slice() : [];
+    const filesName = uploadedFiles.map(file => file.name);
```
jnothman, line 68: This should be sufficient:

```suggestion
    const filesName = (this.state.files || []).map(file => file.name);
```
You don't need to make a copy if you're mapping it and not mutating it..

search/src/Components/upload/react-images-upload/index.js
```diff
@@ -88,6 +91,15 @@ class ReactImageUploadComponent extends React.Component {
         this.setState({fileErrors});
         continue;
       }
+      // Check for duplication
+      if (!this.props.images.includes(file.name) || filesName.includes(file.name)){
+        fileError = Object.assign(fileError, {
+          type: ERROR.FILE_DUPLICATE
```
jnothman, line 97: In the previous logic, this case did not trigger an error, it was just skipped.

ElevnLi: Yeah it's skipped but the user can see an error message saying certain image is duplicate

jnothman: In the existing logic, or only in your changed version?

ElevnLi: In the current version of this branch

jnothman:
Is this now ready for review?

search/src/Components/upload/react-images-upload/index.js
jnothman: We definitely don't want this as an actual error if it will stop us clicking Next...?
But I think it's best not to report it as an error in any case. Better just to skip a file that's already been uploaded.

search/src/Components/upload/react-images-upload/index.js
```diff
@@ -85,55 +87,47 @@ class ReactImageUploadComponent extends React.Component {
...
-
-    Promise.all(allFilePromises).then(newFilesData => {
-      const dataURLs = singleImage?[]:this.state.pictures.slice();
-      const files = singleImage?[]:this.state.files.slice();
-
-      newFilesData.forEach(newFileData => {
-        this.uploadFileToServer(dataURLs, files, newFileData);
+      // Check for duplication
+      if (!this.props.images.includes(file.name) || filesName.includes(file.name)){
+        fileError = Object.assign(fileError, {
```
jnothman, line 95: I note that `Object.assign` is an old way to do `fileError = {fileError, ...{k:v}}`

search/src/Components/upload/react-images-upload/index.js
```diff
@@ -85,55 +87,47 @@ class ReactImageUploadComponent extends React.Component {
...
-        }).then(res => {
-            if(res.status === 200){
-                dataURLs.push(newFileData.dataURL);
-                files.push(newFileData.file);
-                this.setState({pictures: dataURLs, files: files});
-                this.props.handleUploaded(files.map(file => file.name));
-            }
-        })
-    }
+  async uploadFileToServer(newFileData) {
```
jnothman, line 112: Let's make this not async. Too confusing to read. Revert to using axios().then(...)

search/src/Components/upload/react-images-upload/index.js
```diff
@@ -211,7 +205,8 @@ class ReactImageUploadComponent extends React.Component {
   }
 
   renderPreviewPictures() {
-    return this.state.pictures.map((picture, index) => {
+    return this.state.pictures.length <= 100 ?
```
jnothman, line 208: can we make this number a constant?

search/src/Components/upload/react-images-upload/index.js
```diff
@@ -221,7 +216,9 @@ class ReactImageUploadComponent extends React.Component {
           </div>
         </div>
       );
-    });
+    })
+    :
+    <p>No preview for images more than 100</p>
```
jnothman, line 221: Ideally in Material world we use Typography instead of p

search/src/Components/upload/react-images-upload/index.js
```diff
@@ -77,6 +78,7 @@ class ReactImageUploadComponent extends React.Component {
           type: ERROR.NOT_SUPPORTED_EXTENSION
         });
         fileErrors.push(fileError);
+        this.setState({fileErrors});
```
jnothman, line 81: Why do we need this here? Isn't setting the errors after creating all the promises fine?

jnothman (commented):
Since a lot of this PR relates to an error case we probably don't want to treat as an error, and a change from Promise.then to async/await which we don't need, it might be easiest to copy the changes we _do_ want into a new PR.

search/src/Components/upload/react-images-upload/index.js
ElevnLi: This triggers to show error of each image. I think it's the most proper way as you don't want to inform users the format is wrong after the whole dataset being uploaded.

search/src/Components/upload/react-images-upload/index.js
ElevnLi: This was inherited from the original library. I reckon not changing it as long as it's compatible.

search/src/Components/upload/react-images-upload/index.js
jnothman: I think the logic in the existing library has some sense to it: files that match the expected format are uploaded, while those that do not match are not uploaded. Also, given the asynchronous image loading and uploading, the user will be informed at roughly the same time since queuing all the images to load is fairly cheap.
However, your change triggers multiple successive renders, when each error is added, so is not great for responsiveness.

search/src/Components/upload/react-images-upload/index.js
geezacoleman: Seems to work pretty well - received duplicate errors, but the process just continued.

search/src/Components/upload/react-images-upload/index.js
```diff
@@ -87,53 +93,38 @@ class ReactImageUploadComponent extends React.Component {
...
+          data: body,
+          headers: {'Content-Type': 'multipart/form-data', 'X-CSRFToken': Cookies.get('csrftoken') }
+      }).then(res => {
+          if(res.status === 200){
+            const dataURLs = singleImage?[]:this.state.pictures.slice();
+            const files = singleImage?[]:this.state.files.slice();
+            dataURLs.push(newFileData.dataURL);
+            files.push(newFileData.file);
+            this.setState({pictures: dataURLs, files: files});
+            this.props.handleUploaded(files.map(file => file.name));
```
jnothman, line 124: This needs to ensure uniqueness, or syncImageErrorMessage needs to ensure uniqueness. Ideally syncImageErrorMessage (and maybe the setting of `pictures` and `files` in state) would be debounced to avoid updating too rapidly. We can discuss.

search/src/Components/upload/react-images-upload/index.js
```diff
@@ -87,53 +93,38 @@ class ReactImageUploadComponent extends React.Component {
...
-                this.props.handleUploaded(files.map(file => file.name));
-            }
-        })
-    }
+  uploadFileToServer(newFileData) {
+      const {singleImage} = this.props;
+      const body = new FormData()
+      body.append('upload_image', newFileData.file)
+      body.append('upload_id', this.props.upload_id)
+      axios({
```
jnothman, line 112: We should try using a shared HTTP agent that keeps the connection alive to reduce overhead here. For example: https://github.com/axios/axios/issues/1846#issuecomment-434208900 (or use [agentkeepalive.Agent](https://www.npmjs.com/package/agentkeepalive) for simpler configuration of multiple concurrent sockets on the one agent).

search/django/weedid/views.py
```diff
@@ -76,10 +77,10 @@ def upload(request):
         images = []
         file_weedcoco = request.FILES["weedcoco"]
         weedcoco_json = json.load(file_weedcoco)
-        validate(
-            weedcoco_json,
-            schema=request.POST["schema"] if request.POST["schema"] else "coco",
-        )
+        # validate(
```
jnothman, line 80: Is this due to the problem on your local dev machine with the EPPO data having an invalid Zip?

ElevnLi: Yeah I'll revert it back

jnothman (commented):
Omg, it seems I had failed to submit this review, which is a good reason you never responded to it @ElevnLi!!

jnothman:
We can discuss what actually should be done, tomorrow....

search/src/Components/upload/upload_stepper.js
```diff
@@ -352,6 +369,22 @@ class UploadStepper extends React.Component {
                 <Typography className={classes.instructions}>
                     {this.getStepContent(stepName)}
                 </Typography>
+                {stepName === "images" ?
```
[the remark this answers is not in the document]
ElevnLi: I'm having trouble returning array from a functional component. When I try to decouple them, React gives me:
```
Error: Invalid hook call. Hooks can only be called inside of the body of a function component. This could happen for one of the following reasons:
1. You might have mismatching versions of React and the renderer (such as React DOM)
2. You might be breaking the Rules of Hooks
3. You might have more than one copy of React in the same app
See https://fb.me/react-invalid-hook-call for tips about how to debug and fix this problem.
```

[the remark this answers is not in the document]
jnothman: Is this in relation to the code you've currently pushed?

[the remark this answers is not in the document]
ElevnLi: I think I made it work now with the latest commit, but returning array of jsx elements from functional component will render them in a row, not allowing decoupling.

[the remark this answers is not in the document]
jnothman: That doesn't make sense to me... not sure how component structure should affect presentation. But I can't investigate now.

search/yarn.lock
```diff
@@ -7151,6 +7151,11 @@ https-browserify@^1.0.0:
   resolved "https://registry.yarnpkg.com/https-browserify/-/https-browserify-1.0.0.tgz#ec06c10e0a34c0f2faf199f7fd7fc78fffd03c73"
   integrity sha1-7AbBDgo0wPL68Zn3/X/Hj//QPHM=
 
+https@^1.0.0:
+  version "1.0.0"
+  resolved "https://registry.yarnpkg.com/https/-/https-1.0.0.tgz#3c37c7ae1a8eeb966904a2ad1e975a194b7ed3a4"
+  integrity sha1-PDfHrhqO65ZpBKKtHpdaGUt+06Q=
+
```
jnothman, line 7158:
```suggestion
```

search/package.json
```diff
@@ -21,6 +21,7 @@
     "@types/react": "^17.0.0",
     "@types/react-dom": "^17.0.0",
     "helmet": "^4.4.1",
+    "https": "^1.0.0",
```
jnothman, line 24:
```suggestion
```

search/src/Components/upload/react-images-upload/index.js
```diff
@@ -3,7 +3,9 @@ import PropTypes from 'prop-types';
 import './index.css';
 import FlipMove from 'react-flip-move';
 import axios from 'axios';
-import Cookies from 'js-cookie'
+import Cookies from 'js-cookie';
+import http from 'http';
+import https from 'https';
```
jnothman, line 8:
```suggestion
```

jnothman (commented):
I don't think we need https anymore

search/django/weedid/utils.py
```diff
@@ -33,6 +36,19 @@ def store_tmp_image(image, image_dir):
...
 
 
+def store_tmp_image_from_zip(upload_image_zip, image_dir, full_images):
+    if not os.path.isdir(image_dir):
+        mkdir_safely(image_dir)
+    existing_images = os.listdir(image_dir)
+    with tempfile.TemporaryDirectory() as tempdir:
+        ZipFile(upload_image_zip).extractall(tempdir)
+        for dir, _, filenames in os.walk(tempdir):
+            for filename in filenames:
```
jnothman, line 46:
```suggestion
            # FIXME: this should reject a zip upload if two filenames are identical
            for filename in filenames:
```

search/django/weedid/views.py
```diff
@@ -216,6 +217,25 @@ def upload_image(request):
...
 
+def upload_image_zip(request):
+    if not request.method == "POST":
+        return HttpResponseNotAllowed(request.method)
+    user = request.user
+    if not (user and user.is_authenticated):
+        return HttpResponseForbidden("You dont have access to proceed")
+    upload_id = request.POST["upload_id"]
+    upload_image_zip = request.FILES["upload_image_zip"]
+    images = request.POST["images"].split(",")
```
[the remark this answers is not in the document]
jnothman: I'm curious about why this is comma-delimited, but not json-encoded. This will break for filenames with `,` in it.

jnothman (commented):
otherwise LGTM

PR #508, opened 2021-06-22
add gunicorn and static file server for mystatic

### LINKED ISSUE #344: django's runserver should be replaced with gunicorn in production

ElevnLi:
Fixes #344

### DIFF
--- a/search/docker-compose.yml
+++ b/search/docker-compose.yml
@@ -6,7 +6,6 @@ services:
     container_name: reverse_proxy
     depends_on:
         - django
-        - reactivesearch
     build:
         context: ./nginx/nginx
     ports:
@@ -22,6 +21,7 @@ services:
     image: django
     container_name: django
     depends_on:
+      - mystatic
       - elasticsearch
       - db
       - redis
@@ -46,7 +46,8 @@ services:
       - repository_dir:/code/repository
       - upload_dir:/code/upload
       - download_dir:/code/download
-    command: python manage.py runserver 0.0.0.0:8000
+      - mystatic_dir:/code/mystatic
+    command: bash -c "python manage.py makemigrations && python manage.py migrate && python manage.py collectstatic <<<yes && gunicorn core.wsgi -b 0.0.0.0:8000"
     entrypoint: /entrypoint.sh
     ports:
       - "8000:8000"
@@ -195,6 +196,22 @@ services:
     logging:
       driver: "json-file"
 
+  mystatic:
+    image: mystatic
+    container_name: mystatic
+    build:
+      context: ./nginx/mystatic
+    volumes:
+      - mystatic_dir:/mystatic
+    ports:
+      - "8083:80"
+    networks:
+      - elastic
+    stdin_open: true
+    restart: always
+    logging:
+      driver: "json-file"
+
   fluentd:
     container_name: fluentd
     user: root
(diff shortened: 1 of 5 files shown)


search/docker-compose.yml
```diff
@@ -46,7 +46,8 @@ services:
       - repository_dir:/code/repository
       - upload_dir:/code/upload
       - download_dir:/code/download
-    command: python manage.py runserver 0.0.0.0:8000
+      - mystatic_dir:/code/mystatic
+    command: bash -c "python manage.py makemigrations && python manage.py migrate && python manage.py collectstatic <<<yes && gunicorn core.wsgi -b 0.0.0.0:8000"
```
jnothman, line 50:
```suggestion
    command: bash -c "python manage.py makemigrations && python manage.py migrate && python manage.py collectstatic --noinput && gunicorn core.wsgi -b 0.0.0.0:8000"
```

search/docker-compose.yml
```diff
@@ -195,6 +196,22 @@ services:
     logging:
       driver: "json-file"
 
+  mystatic:
+    image: mystatic
+    container_name: mystatic
+    build:
+      context: ./nginx/mystatic
+    volumes:
+      - mystatic_dir:/mystatic
```
jnothman, line 205: It may not be ideal to do this with a volume, but it'll work for now.

jnothman (commented):
It would be nice if we had a single file server that served thumbnails, downloads, static, etc. Is there a reason not to do this??

search/docker-compose.yml
ElevnLi: It might be asking whether to overwrite existing files.

ElevnLi:
> It would be nice if we had a single file server that served thumbnails, downloads, static, etc. Is there a reason not to do this??

It is better to have them together if we want to keep all these contents public and don't need Django to control the access level of say, the dataset download. The main reason we have multiple is that we set them up gradually.
I can combine them into one but it will take some time to migrate, synchronise and test.

jnothman:
> The main reason we have multiple is that we set them up gradually.
I can combine them into one but it will take some time to migrate, synchronise and test.

Test, yes, but I would've thought that replacing `proxy_pass` with `root` here https://github.com/Sydney-Informatics-Hub/Weed-AI/blob/92cf259b224e590df3c31ac6ccfe114251070977/search/nginx/nginx/nginx.conf#L31-L37 and moving volumes directives in docker-compose to the main reverse_proxy service or a single static files service would take about 20 minutes and reduce duplication substantially.

jnothman:
We can merge this PR without it, but I'm not very comfortable with the level of duplication in this repo.

ElevnLi:
> > The main reason we have multiple is that we set them up gradually.
> > I can combine them into one but it will take some time to migrate, synchronise and test.
> 
> Test, yes, but I would've thought that replacing `proxy_pass` with `root` here
> 
> https://github.com/Sydney-Informatics-Hub/Weed-AI/blob/92cf259b224e590df3c31ac6ccfe114251070977/search/nginx/nginx/nginx.conf#L31-L37
> 
> and moving volumes directives in docker-compose to the main reverse_proxy service or a single static files service would take about 20 minutes and reduce duplication substantially.

Testing and debugging to make it work flawlessly could cost couple of hours or more from me. Some seemingly easy issue like the one of protehome may be just not working for some reason and end up with bad alternatives. I reckon to postpone the insignificant matters.

jnothman:
Okay

jnothman (approved):

PR #611, opened 2022-02-09
Cvat uploader as submodule

spikelynch:
This branch now includes the cypress tests for cvat, and I've updated frontend.yml  and prod docker-compose.yml to init the cvat submodule and build it.


spikelynch:
Bother, pr'd this against the wrong branch.

PR #629, opened 2022-02-24
Edit dataset master

ElevnLi:


ElevnLi:
Screen Shot 2022-03-03 at 11 37 17 am

jnothman:
Instead of adding an extra column, how about just adding an [Edit icon](https://mui.com/components/material-icons/?query=edit&selected=Edit) to an existing column (either name or explore column)?

I have also wondered whether this page would be better rendered as a series of MUI Cards. Cards are designed to show things like title, actions, other data.

ElevnLi:
Screen Shot 2022-03-03 at 12 11 40 pm

jnothman (commented):
Looks good. Now I'm wondering how to make it look like it's "edit dataset" rather than "edit title". A tooltip over the button?

jnothman:
Yeah, use a tooltip with "Edit Dataset". Like this example: https://mui.com/components/tooltips/#basic-tooltip

ElevnLi:
Screen Shot 2022-03-03 at 12 29 47 pm

PR #643, opened 2022-03-08
Fix error message of zip file upload

ElevnLi:


ElevnLi:
To fix the error message of zip upload. Currently it saves missing_images in the task record, but actually that list is uploaded_images, and frontend of zip uploader should know how many images still missing.

hlydecker (commented):
Passes checks, doesn't change a lot of code, doesn't look harmful. Time to merge.

PR #652, opened 2022-03-17
Include annotations in the image hash

ElevnLi:


weedcoco/repo/deposit.py
```diff
@@ -123,16 +128,16 @@ def extract(self, dest_dir, version="head"):
             objdir=str(self.object_path), version=version, dstdir=dest_dir
         )
 
-    def validate(self):
+    def validate_image(self):
```
spikelynch, line 131: I think this was better as validate() in case we want to include other checks which aren't to do with image validation

weedcoco/repo/deposit.py
```diff
@@ -223,13 +228,44 @@ def get_all_existing_hash(self, repository):
...
+            }
+            for annotation in weedcoco["annotations"]
+        }
+        for image_name, image_annotation in image_annotation_mapping.items():
+            annotations_hash[image_name] = hashlib.md5(
+                json.dumps(image_annotation, sort_keys=True).encode("utf-8")
+            ).hexdigest()
+        return annotations_hash
+
+    def create_image_hash(self, annotations_hash):
```
spikelynch, line 257: I like that we're keeping the image hash and annotation hash separate - it means we don't lose the ability to check for duplicate images with different annotations

spikelynch (changes requested):

PR #653, opened 2022-03-18
Editable datasets with ocfl repository

spikelynch:
This supersedes the old ocfl PR #633


search/django/weedid/views.py
```diff
@@ -117,11 +137,51 @@ def upload(request):
...
+        return HttpResponseForbidden("You dont have access to proceed")
+    dataset = Dataset.objects.get(upload_id=dataset_id, status="C")
+    if dataset.user.id != user.id:
+        return HttpResponseForbidden("You dont have access to edit")
+    dataset_path = os.path.join(REPOSITORY_DIR, dataset_id)
+    upload_path = os.path.join(UPLOAD_DIR, str(user.id), dataset_id)
+    if os.path.isdir(dataset_path):
+        if os.path.isdir(upload_path):
+            shutil.rmtree(upload_path)
+        shutil.copytree(dataset_path, upload_path)
```
ElevnLi, line 161: I think for this part, we need to retrieve the weedcoco.json and images using the upload_id from the ocfl repo and store them into the upload dir for editing. Previously we just copy them directly from the repo dir when the repo and upload were under the same structure

spikelynch: Yes - that's what I started doing (I haven't pushed that work yet) but it's tricky for two reasons. One is that the image names are different, the other is that checking out the ocfl repo may take a while for large datasets.

What I'm working on now works like this:

* set up an empty upload dir and extract the weedcoco.json into it
* upload any incoming images in to the upload dir
* hash the new images and check for duplicates
* in the celery task, check out a copy of the complete dataset, copy the hashed images and weedcoco.json into the copy, and then commit that as an update

spikelynch: One thing I don't like about ocfl-py is that it doesn't provide incremental updates (ie just change / add / delete one file)

spikelynch: The above will need to be a bit more complex when using the redis store to preserve the image file mapping - that's what I want to do tomorrow

ElevnLi: My thinking of the workflow looks like:
1. When clicking the edit icon of a dataset, we will copy the weedcoco.json file and the images dir into the upload dir with the original name
2. If the user doesn't modify the annotation, then we can keep all the images as what they. However if the user decides to upload new annotations, then we have to make some decisions between below:
a. rename all the images to their original names
b. copy all the images with the original names, so inside the images dir there will be both `5f1218b0e469003cc0a1-986e5d71ce7562a90ff91581a9f2617d.png` and `001_image.png`
c. Add all the original names to a set, which helps generate the list of images that should be uploaded for this dataset
3. Then the users can continue with the stepper, as either way they don't have to upload same images again no matter whether they change the annotation or not.

The reason of having the mapping, which is as shown below, is that we don't want users to re-upload the same image, and I think it's better than letting them upload and then compare the uniqueness, coz if they are going to upload anyway, we can just do overwrite.

```sh
/data # redis-cli
127.0.0.1:6379> get 5f1218b0e469003cc0a1-986e5d71ce7562a90ff91581a9f2617d.png
"32e51313-9d02-4f0c-a11f-0b9ddb6d7c42/001_image.png"
```

spikelynch: Agree that we shouldn't be requiring the user to re-upload images if they already exist.

The other case - which we may not want to handle in the initial version - might be where the user wants to add new images to a dataset. Mostly what we've talked about so far is updating the metadata or agcontext, or adding / updating the annotations to an existing set of images.

spikelynch: Another possible problem just occured to me - if the user downloads a dataset, and adds annotations to that dataset, the image names in the weedcoco.json will be the hashed ones, not the originals. But the editing code can probably detect that (by seeing if the filenames look like HASH1-HASH2.EXT?)

ElevnLi: > Another possible problem just occured to me - if the user downloads a dataset, and adds annotations to that dataset, the image names in the weedcoco.json will be the hashed ones, not the originals. But the editing code can probably detect that (by seeing if the filenames look like HASH1-HASH2.EXT?)

Yeah that's why we copy images with hash_name first and see if user decides to update the annotation. If they do need to touch the annotation then we rename the images using the mapping or keep each image in 2 copies with both hash_name and original_name

hlydecker:
already in main dev branch

PR #667, opened 2022-04-05
Bugfix for ocfl deduping

spikelynch:
Fix for the bug which was making ocfl keep a copy of every image in a dataset in each version. The migrate_images method of RepositoryDataset was stripping exif every time it was run, and this was causing changes to the files even when there was no exif metadata to strip, so ocfl couldn't deduplicate them.

I've moved the exif-stripping code to weedcoco.utils.copy_without_exif, and it only gets called when the files are uploaded via weedid.utils.store_tmp_image and store_tmp_image_from_zip.

There's now a more comprehensive test case for depositing v1 of a dataset with four files, and then v2 with an extra file and its annotations, and the test simulates what happens in the web app where we extract the last head amd then rename the image files from the hashed versions back to the originals. It does this by storing the hash mappings as json fixtures - these can be regenerated like the rest of the sample test fixtures using the --rewrite-deposit-truth flag.


ElevnLi:
Can we rebase this pr to another branch, so that the changes are clearer. Thanks.

PR #669, opened 2022-04-06
Zipfiles for older version of a dataset now visible in the summary page

spikelynch:
They're in a drop-down underneath the yellow downloads button - may need some UX tidying.

### DIFF
--- a/search/src/Components/dataset/dataset_summary.js
+++ b/search/src/Components/dataset/dataset_summary.js
@@ -24,6 +24,9 @@ import IconButton from '@material-ui/core/IconButton';
 import Card from '@material-ui/core/Card';
 import CardContent from '@material-ui/core/CardContent';
 import { makeStyles } from "@material-ui/core/styles";
+import FormControl from '@material-ui/core/FormControl';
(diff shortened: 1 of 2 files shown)


ElevnLi:
Screen Shot 2022-04-07 at 11 03 33 am

search/src/Components/dataset/dataset_summary.js
```diff
@@ -303,6 +323,8 @@ export const DatasetSummary = (props) => {
     const datasetUrl = "https://weed-ai.sydney.edu.au/datasets/" + upload_id;
     displayMeta["identifier"] = [...displayMeta.identifier || [], datasetUrl];
     const authorList = listToWords(displayMeta.creator.map((creator) => creator.name));
+    const head = parseInt(head_version);
+    const old_versions = isNaN(head) ? [] : [...Array(head - 1).keys()].map((i) => i + 1).reverse();
```
ElevnLi, line 327: I thought this was retrieving available previous versions. So this is assuming that we will have all the versions available till the head version.

PR #686, opened 2022-06-09
Last fixes

ElevnLi:


ElevnLi:
fixes #686

search/docker-compose.yml
```diff
@@ -285,8 +284,6 @@ services:
     restart: always
     depends_on:
       - cvat
-    environment:
-      BASENAME: ${CVAT_BASEPATH}
```
spikelynch, line 289: We're taking these out because they weren't needed, is that right? Because the cvat build is happening via this docker-compose, we don't need to pass the path down to its build environment

spikelynch (approved):
Just wanted to check my understanding of the BASENAME removal from docker-compose.ym but it's good to go

